PR #289, opened 2026-04-06
fix: Invalidate root layout on auth state mismatch from prerendered pages

stickerdaniel:
Prerendered marketing pages bake authState.isAuthenticated: false at
build time. When client-side auth recovers from cookies, SvelteKit
skips re-running the root +layout.server.ts (no tracked dependency
change), leaving viewer: null. The {#if user} guard in
AuthenticatedLayout then renders nothing — blank page.

Add event.depends('app:auth') to root layout and detect client/server
auth state divergence in AppAuthProvider via invalidate('app:auth').
Only fires once per page session when states disagree; zero overhead
for normal navigation.

### DIFF
--- a/src/lib/components/app/app-auth-provider.svelte
+++ b/src/lib/components/app/app-auth-provider.svelte
@@ -1,6 +1,8 @@
 <script lang="ts">
 	import { page } from '$app/state';
-	import { createSvelteAuthClient } from '@mmailaender/convex-better-auth-svelte/svelte';
+	import { invalidate } from '$app/navigation';
+	import { browser } from '$app/environment';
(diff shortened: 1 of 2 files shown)


stickerdaniel:
* **#289** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/289?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/289?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #313, opened 2026-04-12
feat(oxlint): Add no-invalid-module-path rule for Convex modules

### LINKED ISSUE #309: Add static check for invalid Convex module filenames
## Problem

Cloudflare preview builds can fail during `bun scripts/deploy.ts` when Convex rejects a module path that passed local development and type checks.

Concrete failure:

```text
InvalidConfig: aiChat/agent-runtime.js is not a valid path to a Convex module.
Path component agent-runtime.js can only contain alphanumeric characters, underscores, or periods.
```

In this case, a file under `src/lib/convex/**` used a hyphen in its basename (`agent-runtime.ts`). `bun run check:convex` did not catch it, and the failure only surfaced during `convex deploy` inside the Workers build.

## Why this matters

This is a template-level footgun:
- it causes CI/deploy failures late in the pipeline
- the error is easy to trigger accidentally
- the allowed path-component rule is stricter than many developers will expect

## Proposed fix

Add a guard to `scripts/static-checks.ts` that scans `src/lib/convex/**/*.ts` and fails if any path component under `src/lib/convex/` does not match Convex's module-path constraints.

Suggested rule:
- allow only `^[A-Za-z0-9_.]+$` for each path component
- reject filenames like `agent-runtime.ts`
- allow filenames like `agentRuntime.ts`, `agent_runtime.ts`, and `agent.runtime.ts`

## Why `static-checks.ts`

This is primarily a filesystem/deploy invariant, so the shared static-check pipeline is the most reliable place to enforce it:
- pre-commit catches it before push
- CI catches it before deploy
- applies even if the file is not imported yet

Oxlint could mirror this later, but `static-checks.ts` should be the canonical backstop.

## Acceptance criteria

- `bun scripts/static-checks.ts --staged` fails on an invalid Convex module path
- error output points to the offending file/path component
- valid Convex module names continue to pass unchanged

stickerdaniel:
Validate Convex module filenames against the backend's path rules
(alphanumeric, underscore, period only; at least one alphanumeric).
Catches deploy-breaking filenames like agent-runtime.ts at lint time.

Resolves: #309

### DIFF
--- a/.oxlintrc.json
+++ b/.oxlintrc.json
@@ -9,6 +9,7 @@
 				"ignorePatterns": ["tests.*", "localDev.*"],
 				"ignoreUsageFiles": ["**/*.test.ts", "e2e/**"]
 			}
-		]
+		],
+		"convex/no-invalid-module-path": "error"
 	}
 }
--- a/patches/oxlint-plugin-convex@0.1.1.patch
+++ b/patches/oxlint-plugin-convex@0.1.1.patch
@@ -1,5 +1,8 @@
+diff --git a/node_modules/oxlint-plugin-convex/.bun-tag-c2a66dab253f6d5a b/.bun-tag-c2a66dab253f6d5a
+new file mode 100644
+index 0000000000000000000000000000000000000000..e69de29bb2d1d6434b8b29ae775ad8c2e48c5391
 diff --git a/dist/index.mjs b/dist/index.mjs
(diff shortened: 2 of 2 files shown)


stickerdaniel:
* **#313** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/313?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/313?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #314, opened 2026-04-12
fix(env): Replace $env/dynamic/public with $env/static/public

### LINKED ISSUE #303: fix: PostHog and snapdom use $env/dynamic/public unnecessarily
## Summary

Three files in the template use `$env/dynamic/public` for values that would be better served by `$env/static/public`. This is a minor footgun that cargo-culted across the codebase.

## Background

SvelteKit exposes env vars through four modules:

- `$env/static/public` — read at **build time**, baked into the bundle, tree-shakable, build-time validated
- `$env/dynamic/public` — read at **runtime**, fetched from platform env, **platform-adapter dependent**
- `$env/static/private` — server-only build-time
- `$env/dynamic/private` — server-only runtime

The Svelte docs and platform guidance both recommend **static by default** — only go dynamic when you genuinely need runtime resolution (e.g., one build → many environments, platform-injected feature flags, etc.).

On Cloudflare Workers specifically, `$env/dynamic/public` behavior depends on how the adapter exposes runtime env — static is more reliable.

## Affected files

| File | Variable | Current | Recommended |
|---|---|---|---|
| `src/lib/analytics/posthog.ts` | `PUBLIC_POSTHOG_API_KEY`, `PUBLIC_POSTHOG_HOST`, `PUBLIC_POSTHOG_PROXY_HOST` | `$env/dynamic/public` | `$env/static/public` |
| `src/lib/components/app/app-posthog-bootstrap.svelte` | `PUBLIC_POSTHOG_API_KEY`, `PUBLIC_POSTHOG_HOST` | `$env/dynamic/public` | `$env/static/public` |
| `src/lib/utils/snapdom-config.ts` | `PUBLIC_SNAPDOM_PROXY_URL` | `$env/dynamic/public` | `$env/static/public` |

Sentry already uses `$env/static/public` correctly (PR #302) — good reference.

## Why these should be static

**PostHog keys** — never change per request, set once per deployment. Same project key for all CF Workers Builds of the same environment.

**snapDOM proxy URL** — deployment-time config for CORS proxy. Not a runtime feature flag.

**No runtime variability requirement** — these are all set via `varlock`/`.env.local`/CF Workers build variables, all available at build time.

## Benefits of switching

1. **Build-time validation** — missing var fails the build loudly instead of silently returning `undefined` at runtime
2. **Dead-code elimination** — `if (!PUBLIC_POSTHOG_API_KEY) return null` gets stripped entirely when the key is empty, removing the whole PostHog init path from the bundle
3. **No runtime lookup overhead** — values become string literals
4. **TypeScript narrowing** — named imports give better type inference than object access
5. **Consistency with Sentry** — template should demonstrate one canonical pattern

## Example migration

```ts
// src/lib/analytics/posthog.ts — BEFORE
import { env } from '$env/dynamic/public';
// ...
const apiKey = env.PUBLIC_POSTHOG_API_KEY;
const host = env.PUBLIC_POSTHOG_HOST;

// AFTER
import { PUBLIC_POSTHOG_API_KEY, PUBLIC_POSTHOG_HOST, PUBLIC_POSTHOG_PROXY_HOST } from '$env/static/public';
// ...
if (!PUBLIC_POSTHOG_API_KEY || !PUBLIC_POSTHOG_HOST) return null;
```

## When dynamic WOULD be justified

Keep `$env/dynamic/*` for genuine runtime-only concerns:

- `src/lib/server/convex-http.ts` — `CONVEX_INTERNAL_URL` is set by deploy script at runtime, dynamic is correct
- `src/routes/[[lang]]/emails/[...email]/+page.server.ts` — dev-only dynamic import pattern for `RESEND_API_KEY`, correct

## Notes

Found while porting PR #302 (Sentry integration) downstream. Initial port correctly used static for Sentry, but an AI reviewer (Copilot) flagged the client import as a "server/client mismatch" because `hooks.server.ts` was already using dynamic (also wrong). The real fix is making them all static.

stickerdaniel:
## Summary

- Migrate PostHog and snapdom from `$env/dynamic/public` to `$env/static/public` — enables dead-code elimination (optional integrations stripped when unconfigured), build-time validation, and avoids `/_app/env.js` waterfall on prerendered pages
- Add ESLint `no-restricted-imports` rule to prevent `$env/dynamic/public` from being reintroduced

## Test plan

- [x] `bun scripts/static-checks.ts` passes on all changed files
- [x] `bun run build` succeeds (static imports compile)
- [x] `grep -r '$env/dynamic/public' src/` returns zero matches
- [ ] Verify PostHog initializes correctly on preview deploy (when keys are set)

Resolves #303

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -124,6 +124,11 @@ export default defineConfig(
 						{
 							name: '$lib/utils/utils.js',
 							message: 'Import from $lib/utils.js instead (canonical location).'
+						},
+						{
+							name: '$env/dynamic/public',
+							message:
+								'Use $env/static/public instead. All PUBLIC_* vars are known at build time in this project.'
 						}
 					]
 				}
--- a/src/lib/analytics/posthog.ts
+++ b/src/lib/analytics/posthog.ts
@@ -1,4 +1,8 @@
-import { env } from '$env/dynamic/public';
+import {
+	PUBLIC_POSTHOG_API_KEY,
+	PUBLIC_POSTHOG_HOST,
+	PUBLIC_POSTHOG_PROXY_HOST
+} from '$env/static/public';
 
 const READY_EVENT = 'posthog:ready';
 const ADBLOCK_DETECT_TIMEOUT_MS = 3000;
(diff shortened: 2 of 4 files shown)


stickerdaniel:
* **#314** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/314?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/314?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #319, opened 2026-04-12
feat(codegen): Add smart generate wrapper for local Convex backends

### LINKED ISSUE #312: Improve Convex codegen UX for local embedded backend projects
## Problem

In this template, local development normally runs against the embedded local Convex backend, and `.env.local` is often set to something like:

```env
CONVEX_DEPLOYMENT=local:local-...
PUBLIC_CONVEX_URL=http://127.0.0.1:3210
```

That works fine for app development, but `bun run generate` currently runs:

```bash
bun run convex codegen
```

When no cloud deployment is configured, this fails with Convex's default message:

```text
✖ Failed to load deployment config - try running `npx convex dev --configure`
```

## Why this is a problem

That message is a poor fit for the template's default local-dev setup:

- the project is intentionally using the embedded local backend
- the user's config is not necessarily broken
- the suggested fix points them toward cloud setup even when they may not want or need a cloud deployment yet

This makes `bun run generate` feel broken or misleading for local-only users.

## Repro

1. Use the template in local embedded Convex mode
2. Set `.env.local` with a local deployment value like `CONVEX_DEPLOYMENT=local:...`
3. Run:

```bash
bun run generate
```

Current result:

```text
✖ Failed to load deployment config - try running `npx convex dev --configure`
```

## Recommended fix

Wrap codegen in a repo-level script that checks whether a usable cloud deployment is configured before calling Convex codegen.

Suggested behavior:

- if `CONVEX_DEPLOYMENT` is missing or starts with `local:`
  - fail fast with a template-specific message, e.g.
  - `Convex codegen requires a configured cloud deployment; the local embedded backend is not sufficient for codegen in this template.`
- optionally split commands into something like:
  - `bun run generate` for safe local generation tasks
  - `bun run generate:convex` or `bun run generate:cloud` for Convex codegen

## Benefit

This would prevent confusion for users following the template's default local development flow and make the failure mode much more understandable.

stickerdaniel:
## Summary

- Wraps `bun run generate` in a smart script that auto-detects the Convex environment (self-hosted, cloud, local embedded backend, or offline) and runs the appropriate codegen strategy
- Fails early with clear, template-specific error messages instead of the confusing default `try running npx convex dev --configure`
- When offline with existing generated types, validates them in-place instead of destructively overwriting with incomplete stubs
- Detects conflicting/partial env var configurations (e.g. both `CONVEX_DEPLOYMENT` and `CONVEX_SELF_HOSTED_URL` set)

Resolves #312

## Test plan

- [ ] With `bun run dev` running: `bun run generate` detects local backend and runs full codegen
- [ ] With dev server stopped and `_generated/` present: validates existing types offline
- [ ] With cloud `CONVEX_DEPLOYMENT` set: passes through to `convex codegen`
- [ ] With conflicting env vars: fails early with clear conflict message
- [ ] With partial self-hosted config: fails early with incomplete config message

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -100,7 +100,7 @@ Local dev notes (`bun run dev`):
 
 **IMPORTANT:** When any task involves Convex backend code — writing, reviewing, or modifying queries, mutations, actions, schema, HTTP endpoints, auth, file storage, or crons — you MUST read the `convex-guidelines` skill (`skills/convex-guidelines/SKILL.md`) first. It contains the canonical Convex coding patterns for this project.
 
-- `bun run generate` - Regenerate Convex type definitions (`_generated/`). Auto-detects the environment: uses cloud/self-hosted deployment if configured, local embedded backend if `bun run dev` is running, or validates existing types offline. Safe to run in any environment.
+- `bun run generate` - Regenerate Convex type definitions (`_generated/`). Auto-detects environment in priority order: cloud/self-hosted deployment (if configured) > local embedded backend (if `bun run dev` is running) > offline validation of existing types. Safe to run in any environment.
 - `bun run check:convex` - Run the Convex TypeScript project check. Run this whenever you change `src/lib/convex/**` or shared code imported by Convex.
 
 - `bun convex env set KEY value` - Set Convex environment variables (cloud)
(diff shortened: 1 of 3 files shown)


stickerdaniel:
* **#319** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/319?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/319?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #338, opened 2026-04-30
fix(convex): Materialize active streams only

### LINKED ISSUE #334: fix: messageListing materializes finished streams over historical page, stranding tool parts at input-streaming
## Problem

`src/lib/convex/support/messageListing.ts:listMessagesForThread` calls `syncStreams` with `includeStatuses: ['streaming', 'finished', 'aborted']`, then forwards **all** returned streams into `mergeRecentStreamsIntoPage`:

```ts
const materializedPage = await mergeRecentStreamsIntoPage(ctx, {
    threadId: args.threadId,
    page: enrichedPage,
    streamMessages: streams.messages ?? []   // includes finished/aborted
});
```

`mergeAssistantMessage` then unconditionally replaces `parts` on the historical message with the materialized streamed parts:

```ts
return { ...message, status: materialized.status, text: materialized.text,
         parts: materialized.parts as ChatMessage['parts'], ... };
```

For a long-running tool whose deltas don't end with a clean `output-available` transition, the historical `MessageDoc` (which `toUIMessages` correctly resolves to `state: 'output-available'` once the tool result is saved) gets clobbered by the in-flight stream snapshot. The UI then renders the tool forever in `state: 'input-streaming'`, even though the result is durably saved in Convex.

## Repro

1. Add any tool whose UI conditions on `state === 'output-available'` (e.g. inline UI render, structured output card).
2. Send a message that triggers it.
3. Wait for the stream to finish — Convex `agent` `messages` table shows the tool-result row, `streams` row goes `finished`.
4. Reload the page. The tool part still shows the streaming/loading variant. The downstream assistant text after the tool also goes missing.

We hit this with a `renderUI` tool (json-render shadcn catalog) on a hackathon fork. DB had `{ok: true, spec: ...}` saved; UI stayed on "Processing tool call..." indefinitely.

## Fix

Only materialize streams that are still `streaming`. Once `finished`/`aborted`, the persisted MessageDoc is authoritative — `toUIMessages` already produces the correct `output-available` shape.

```ts
const activeStreamMessages = (streams.messages ?? []).filter(
    (s) => s.status === 'streaming'
);
const materializedPage =
    activeStreamMessages.length === 0
        ? enrichedPage
        : await mergeRecentStreamsIntoPage(ctx, {
              threadId: args.threadId,
              page: enrichedPage,
              streamMessages: activeStreamMessages
          });
const liveStreams = { kind: 'list' as const, messages: activeStreamMessages };
```

Repo where it was first hit: `stickerdaniel/hackUPC` — `web/src/lib/convex/support/messageListing.ts:93-106`.

stickerdaniel:
syncStreams returns streaming, finished, and aborted streams when called
with includeStatuses set to all three. listMessagesForThread was passing
the full list into mergeRecentStreamsIntoPage, so finished streams
clobbered the persisted MessageDoc (already at output-available) with
their in-flight snapshot. Tools whose final delta did not transition
cleanly stayed at input-streaming after reload, and any text following
the tool part went missing.

Filter to status === 'streaming' before materialization. The persisted
rows are authoritative once a stream finishes. Reuse the filtered array
for liveStreams so the response stays consistent.

Closes #334

### DIFF
--- a/src/lib/convex/support/messageListing.ts
+++ b/src/lib/convex/support/messageListing.ts
@@ -90,17 +90,27 @@ export async function listMessagesForThread(
 		return { ...paginated, page: enrichedPage, streams };
 	}
 
-	const materializedPage = await mergeRecentStreamsIntoPage(ctx, {
-		threadId: args.threadId,
-		page: enrichedPage,
-		streamMessages: streams.messages ?? []
-	});
+	// Only materialize ACTIVE streams over the historical page. Once a stream
+	// is finished/aborted, the persisted MessageDoc rows are authoritative
+	// (toUIMessages already sets state='output-available' for completed tool
+	// calls). Materializing finished streams was clobbering that with the
+	// in-flight stream snapshot, leaving renderUI stuck at input-streaming
+	// even though the result was saved.
+	const activeStreamMessages = (streams.messages ?? []).filter(
+		(streamMessage) => streamMessage.status === 'streaming'
+	);
+	const materializedPage =
+		activeStreamMessages.length === 0
+			? enrichedPage
+			: await mergeRecentStreamsIntoPage(ctx, {
+					threadId: args.threadId,
+					page: enrichedPage,
+					streamMessages: activeStreamMessages
+				});
 
 	const liveStreams = {
 		kind: 'list' as const,
-		messages: (streams.messages ?? []).filter(
-			(streamMessage) => streamMessage.status === 'streaming'
-		)
+		messages: activeStreamMessages
 	};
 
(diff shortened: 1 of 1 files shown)


stickerdaniel:
* **#338** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/338?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/338?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #339, opened 2026-04-30
fix(prompt-suggestion): truncate long labels and use rounded-lg

### LINKED ISSUE #335: AI chat suggestion chips overflow when label is long — needs truncation
## Bug

The empty-state suggestion chips in the AI chat (`PromptSuggestion` rendered inside `src/routes/[[lang]]/app/ai-chat/thread-chat.svelte`) do not truncate long labels, so any chip whose text exceeds the row width forces the entire chip group to stack one chip per line and overflow off the right edge of the viewport.

## Repro

Replace the default suggestion text with a long sentence, e.g.

```svelte
const suggestions = [
  { text: 'Run a barcelona-baseline scenario for one year (52 ticks). I want to see how the printer ages.', label: '...' },
  { text: 'On my latest run, find the component that degraded fastest and use inspectSensorTrust to tell me whether the sensor reading is trustworthy.', label: '...' },
  { text: 'Spawn a phoenix-aggressive run with the same seed as my last barcelona-baseline, then compare the heater’s final-tick health between the two and explain the gap.', label: '...' }
];
```

The chips render full-width and the third one runs off the right edge of the chat column. Even on a 1440px viewport none of the three chips share a row.

## Expected

- Each chip caps at a sensible max-width (e.g. `14rem` / `max-w-xs`).
- Long labels truncate with an ellipsis (`text-overflow: ellipsis`).
- Full text remains discoverable on hover (`title` attribute) and is still sent verbatim into the textarea on click.
- Three short chips still share a row on desktop; long ones still wrap cleanly on narrow viewports.

## Suggested fix

Two layers — one in the consumer (`thread-chat.svelte`) for now, and ideally one inside the `PromptSuggestion` component so every consumer benefits.

**Consumer (immediate):**

```svelte
{#each suggestions as suggestion, i (suggestion.text)}
  <div
    class=\"min-w-0 max-w-full flex-1 sm:max-w-[14rem]\"
    title={suggestion.text}
  >
    <PromptSuggestion class=\"w-full min-w-0 max-w-full\" onclick={...}>
      <span class=\"block min-w-0 max-w-full truncate\">{suggestion.label}</span>
    </PromptSuggestion>
  </div>
{/each}
```

**Component (recommended):**

`PromptSuggestion` (shadcn-svelte `Button` wrapper) currently renders children straight into a `Button` that has `whitespace-nowrap`. Without `min-w-0` on the Button and a truncating wrapper around the label, content overflows the parent. Either add a `truncate` prop or wrap children in `<span class=\"block min-w-0 truncate\">...</span>` when text-only children are detected.

## Why it matters

Any AI chat in a real product will eventually need long, demo-friendly prompts (multi-step instructions, RAG queries, scenario specs). The chip layout silently breaking on long content is a footgun for downstream forks.

## Environment

- Discovered in a downstream fork on Svelte 5 / SvelteKit 2 / Tailwind v4
- Tailwind utilities used in suggested fix (`truncate`, `min-w-0`, `max-w-[14rem]`) are stock v4

### LINKED ISSUE #336: chips using rounded full instead of adaptice rounded lg

stickerdaniel:
Non-highlight chips hardcoded rounded-full and lacked min-w-0, so
long labels broke flex-wrap layout and the radius looked oversized
on wide chips. Switch to rounded-lg, add min-w-0 to the Button, and
add an opt-in `truncate` prop that wraps text children in a
truncating span. Apply truncate + max-width + title in the AI chat
empty-state suggestions and ChatInput suggestions.

Resolves: #335
Resolves: #336

### DIFF
--- a/scripts/static-checks.ts
+++ b/scripts/static-checks.ts
@@ -293,6 +293,8 @@ async function main(): Promise<void> {
 					formatFlag,
 					'--plugin',
 					'prettier-plugin-svelte',
+					'--plugin',
+					'prettier-plugin-tailwindcss',
 					...formattableFiles
 				]);
 			} else if (!scopedMode) {
--- a/src/lib/chat/ui/ChatInput.svelte
+++ b/src/lib/chat/ui/ChatInput.svelte
@@ -214,7 +214,7 @@
 				<div class="flex flex-wrap gap-2">
 					{#each suggestions as suggestion, i (suggestion.text)}
 						<div
-							class="min-w-0 max-w-full motion-safe:animate-[chip-in_375ms_ease-out_both] sm:max-w-[14rem]"
+							class="max-w-full min-w-0 motion-safe:animate-[chip-in_375ms_ease-out_both] sm:max-w-[14rem]"
 							style="animation-delay: {i * 50}ms"
 							title={suggestion.text}
 						>
--- a/src/lib/chat/ui/ChatInput.svelte
+++ b/src/lib/chat/ui/ChatInput.svelte
@@ -214,10 +214,15 @@
 				<div class="flex flex-wrap gap-2">
 					{#each suggestions as suggestion, i (suggestion.text)}
 						<div
-							class="motion-safe:animate-[chip-in_375ms_ease-out_both]"
+							class="min-w-0 max-w-full motion-safe:animate-[chip-in_375ms_ease-out_both] sm:max-w-[14rem]"
 							style="animation-delay: {i * 50}ms"
+							title={suggestion.text}
 						>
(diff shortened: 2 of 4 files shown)


stickerdaniel:
* **#339** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/339?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/339?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #340, opened 2026-04-30
fix(scripts): sanitize git env in pre-commit hook

### LINKED ISSUE #332: static-checks.ts: prefix detection and re-stage break under pre-commit framework in monorepo subdir
## Context

`scripts/static-checks.ts --staged` (the pre-commit entrypoint) silently corrupts commits when **both** of these conditions are true:

1. The SvelteKit app is nested in a subdirectory of the git repo (e.g. a monorepo with the saas-starter app under `web/` next to other packages), AND
2. The repo uses the [`pre-commit` framework](https://pre-commit.com/) (Python tool) instead of Husky to dispatch hooks.

The combination is uncommon today (Husky + repo-root layout works fine), but neither condition is unreasonable, and the failure mode is severe.

## Symptoms

When committing in this configuration:

1. `prettier` errors with `No files matching the pattern were found: \"web/components.json\"` (and one such line per staged file).
2. If the user retries by clearing things and committing again, the commit \"succeeds\" but `git show --stat HEAD` shows **only deletions** — every staged file is recorded as deleted from the worktree subdir, so the next pull/push wipes them.

## Root cause

The `pre-commit` framework runs each hook with `GIT_DIR` set to the worktree's gitdir (and clears `GIT_WORK_TREE`) so each hook sees an isolated index. With `GIT_DIR` set explicitly:

- `git rev-parse --show-prefix` returns empty even when run from a subdirectory.
- `git rev-parse --show-toplevel` returns `process.cwd()` (so any cwd is treated as the toplevel).

Two places in `scripts/static-checks.ts` rely on these values:

1. **`getStagedFiles()`** uses `--show-prefix` to strip the monorepo prefix from `git diff --cached` output. Under pre-commit it returns empty, so paths keep the `web/` prefix and get passed to prettier from `cd web` cwd → \"no files matching\".
2. **The `git add ...` re-stage step** at the end runs from `cd web` cwd. Under pre-commit's `GIT_DIR` override, those cwd-relative path arguments end up as bogus root-level index entries that `git status` shows as `AD`, and a subsequent commit emits the staged subdir files as deletions.

## Suggested fix

I have a working patch I'm happy to PR if it looks reasonable:

1. Extract a `getCwdPrefix()` helper that strips `GIT_DIR` / `GIT_WORK_TREE` / `GIT_INDEX_FILE` / `GIT_OBJECT_DIRECTORY` before running `git rev-parse`, with a `path.relative(toplevel, cwd)` fallback.
2. Skip the trailing `git add` re-stage when running under pre-commit (`GIT_DIR` set externally, or `PRE_COMMIT*` env vars present). The pre-commit framework already detects working-tree modifications and stages them itself, so the script's own re-stage is redundant — and harmful — in that context.

Tested against a real monorepo + pre-commit setup; commits now land with the correct file modifications. Happy to open a PR if you'd like.

## Repro

`saas-starter` itself runs the script from repo root, so a minimal repro needs a host repo:

1. Create a parent repo with the saas-starter app under `web/` (any sibling layout works).
2. Add a root-level `.pre-commit-config.yaml` that runs the bundled web hooks (`bash -c 'cd web && bun run static-checks:staged'`).
3. Stage a change in `web/...` and commit.

Environment: macOS 25.4.0, git 2.x, pre-commit 4.x, bun 1.x.

stickerdaniel:
## Summary

Resolves #332. `scripts/static-checks.ts --staged` silently corrupted commits when the saas-starter app was nested in a monorepo subdirectory AND the host repo dispatched hooks via the [pre-commit](https://pre-commit.com/) framework instead of Husky.

`pre-commit` sets `GIT_DIR` to the parent repo's gitdir and clears `GIT_WORK_TREE` per hook. With those env vars set externally:

- `git diff --cached --name-only` returned **all parent-repo staged files** — siblings outside the app dir leaked through to prettier/eslint/misspell.
- `git add` wrote into the wrong index, recording the staged files as deletions. The "successful" commit shipped only deletions; a subsequent push wiped the files.

## Fix

- New `scripts/git-context.ts` helper exports `sanitizedGitEnv()`, `isUnderPreCommit()`, and `getStagedFiles()`. Production and tests import the same exported function — no test/production drift.
- `getStagedFiles()` runs `git diff --cached --name-only --diff-filter=ACMR --relative` with sanitized env. `--relative` filters siblings AND strips the cwd prefix in one shot.
- Re-stage step in `static-checks.ts` now passes sanitized env so `git add` targets the correct index. A one-line hint fires only when the pre-commit framework is detected (via `PRE_COMMIT*` env vars — not `GIT_DIR`, since git itself sets that for every hook), explaining its mandatory abort-then-retry flow.

Husky users see no behavioral change.

## Test plan

- [x] `scripts/git-context.test.ts` — env scrubbing + framework detection (7 cases, including a negative case that asserts `GIT_DIR` alone does not flip the detector).
- [x] `scripts/git-context.staged-files.test.ts` — integration test: real temp git repo, `web/src/a.ts` and `other/b.ts` staged, subprocess imports `getStagedFiles` from the helper, runs with `cwd=tmp/web` and externally-set `GIT_DIR`. Asserts the result is exactly `['src/a.ts']` (sibling filtered, prefix stripped).
- [x] `bun scripts/static-checks.ts` on the four touched files — green.

### DIFF
--- a/scripts/git-context.staged-files.test.ts
+++ b/scripts/git-context.staged-files.test.ts
@@ -1,0 +1,77 @@
+import { spawnSync } from 'child_process';
+import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from 'fs';
+import { tmpdir } from 'os';
+import { join, resolve } from 'path';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+
+// Resolved from the project root (where vitest runs). Avoids `import.meta.url`,
+// which vitest may rewrite to a non-file scheme during transform.
+const HELPER = resolve(process.cwd(), 'scripts/git-context.ts');
+
+/**
+ * Spawn a subprocess that imports the real `getStagedFiles` helper and
+ * prints its result as JSON. Importing the same exported function
+ * `static-checks.ts` consumes ensures the test catches regressions in
+ * the helper without re-implementing the git command.
+ */
+function runGetStagedFiles(cwd: string, extraEnv: Record<string, string> = {}): string[] {
+	const code = `import(${JSON.stringify(HELPER)}).then((m) => { process.stdout.write(JSON.stringify(m.getStagedFiles())); });`;
+	const result = spawnSync('bun', ['-e', code], {
+		cwd,
+		env: { ...process.env, ...extraEnv },
+		encoding: 'utf-8'
+	});
+	if (result.status !== 0) {
+		throw new Error(`harness failed (status ${result.status}): ${result.stderr}`);
+	}
+	return JSON.parse(result.stdout || '[]') as string[];
+}
+
+function gitInTmp(tmp: string, args: string[]): void {
+	const result = spawnSync('git', args, { cwd: tmp, encoding: 'utf-8' });
+	if (result.status !== 0) {
+		throw new Error(`git ${args.join(' ')} failed: ${result.stderr}`);
+	}
+}
+
+describe('getStagedFiles (integration)', () => {
+	let tmp: string;
+
+	beforeEach(() => {
+		tmp = mkdtempSync(join(tmpdir(), 'git-context-test-'));
+
+		gitInTmp(tmp, ['init', '-q', '-b', 'main']);
+		gitInTmp(tmp, ['config', 'user.email', 'test@example.com']);
+		gitInTmp(tmp, ['config', 'user.name', 'Test']);
+		gitInTmp(tmp, ['config', 'commit.gpgsign', 'false']);
+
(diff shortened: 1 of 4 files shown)


stickerdaniel:
* **#340** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/340?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/340?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #350, opened 2026-05-03
feat(chat): relax input size cap for images, harden type gate

stickerdaniel:
## Summary

The 5 MB pre-upload size cap in `handleFilesAdded` / `handlePaste` was rejecting phone photos that `processImage` would shrink to a few hundred KB. The server already enforces the 5 MB cap on the actual stored blob (`src/lib/convex/{support,aiChat}/files.ts`), so the input gate was only blocking otherwise-fixable uploads for image-typed files.

Four-part change in `src/lib/chat/ui/ChatInput.svelte`:

1. **Type-aware input cap** — non-image files keep the 5 MB gate (they upload as-is). Images get a 50 MB absurdity ceiling (`MAX_INPUT_IMAGE_SIZE`) sized to prevent OOM in `createImageBitmap` on mobile, not as a policy cap. New constants are re-exported alongside `MAX_FILE_SIZE` from both `chat/core/index.ts` and `chat/index.ts`.
2. **MIME allowlist on `handleFilesAdded`** — drag/drop bypasses the file picker's `accept` filter today, so `image/heic`/`image/tiff` slip past. New `isAllowedKind` helper trusts non-generic MIME types and only falls back to the file extension when `File.type` is `''` or `'application/octet-stream'` — so `image/heic` named `photo.jpg` stays rejected.
3. **`normalizeMime`** for empty/generic `File.type` (macOS Finder drags) — wraps the file in a new `File` with the inferred MIME so `attachFile` routes images through `processImage` and `FileUploader` doesn't fall back to `application/octet-stream` (which the server would reject).
4. **Tightened post-process fallback** in `attachFile` — when the encoded WebP exceeds `MAX_FILE_SIZE` and the original is also over the server cap, throw a translated error so `ChatContext`'s catch surfaces it before any bytes leave the browser. Was safe under the old 5 MB input gate; not safe under the new 50 MB cap.

i18n: two new keys (en/de/es/fr): `chat.error.file_type_not_allowed`, `chat.error.image_compression_exceeded`.

## Test plan

- [ ] CI green
- [ ] Drop or paste a 30 MB JPEG → attaches, processes, uploads as ~500 KB WebP
- [ ] Pick or paste a 6 MB PDF → rejected with "5MB" toast
- [ ] Drop or paste a 60 MB image → rejected with "50MB" toast
- [ ] Drop an `image/heic` file → rejected client-side with "file type not allowed" toast
- [ ] Drop a `.png` from Finder (often empty `file.type`) → passes via extension fallback, processes, uploads
- [ ] Pathological 6 MB PNG that re-encodes >5 MB AND original >5 MB → client-side throw, placeholder removed, translated toast — no upload


stickerdaniel:
* **#350** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/350?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/350?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #361, opened 2026-05-10
fix(breadcrumb): render all intermediate route segments

### LINKED ISSUE #360: authenticated-header breadcrumb skips intermediate route segments
The breadcrumb in `src/lib/components/authenticated/authenticated-header.svelte` only renders the root crumb and the last URL segment, dropping every segment in between. On a route like `/en/app/settings/sessions` the breadcrumb shows `App › Sessions` instead of `App › Settings › Sessions`.

The cause is in the `breadcrumbs` derivation: when the path has more than two segments past the language/prefix, the code reads `segments[segments.length - 1]` and pushes a single intermediate item for it, ignoring the rest. There is no loop over `segments.slice(prefixSegmentIndex + 1, -1)`, so any nested route deeper than two levels collapses.

Suggested fix: iterate every segment after the prefix, push one breadcrumb item per segment (formatted via the existing kebab-to-Title transform), and mark only the final segment with `isLast: true`. Each item's `href` should be the cumulative path up to that segment, so intermediate crumbs link to the parent page rather than to the current full URL.

Reproduction: navigate to any route nested two or more levels under the app prefix (e.g. `/en/app/settings/sessions`, `/en/app/settings/profile`, `/en/app/teams/<slug>/members`).

Happy to send a PR if useful.

stickerdaniel:
Closes #360

The authenticated header breadcrumb only rendered the root crumb plus the leaf segment, dropping every segment in between. So `/en/app/settings/sessions` showed `App › Sessions` instead of `App › Settings › Sessions`. The leaf also linked to the full current URL instead of its own cumulative path, so once intermediates existed they would have pointed at the wrong place too. No shipped route is currently deeper than two levels under `/app` or `/admin`, but this blocks every planned nested route (e.g. `/app/settings/sessions`, `/admin/users/<id>/members`).

Extract the derivation into a pure helper `buildBreadcrumbs(pathname, routePrefix, rootLabel, lang)` that iterates every segment after the prefix and builds cumulative `href`s, so intermediates link to parent pages. The helper takes `lang: string | undefined` (matches `[[lang]]` `RouteParams`) and normalizes via `getLanguage(lang).code` before passing to the explicit `localizedHref(path, lang)` overload, so it stays free of ambient `$app/state`. Cover the behavior with unit tests for mismatched prefix, root path, kebab-case formatting, missing/non-default language, and 3- and 4-level paths with cumulative hrefs.

`bun run test:unit -- breadcrumbs` passes (10/10). `bun scripts/static-checks.ts` clean for the touched files.

### DIFF
--- a/src/lib/components/authenticated/authenticated-header.svelte
+++ b/src/lib/components/authenticated/authenticated-header.svelte
@@ -7,7 +7,7 @@
 	import LanguageSwitcher from '$lib/components/LanguageSwitcher.svelte';
 	import LightSwitch from '$lib/components/ui/light-switch/light-switch.svelte';
 	import CommandTrigger from '$lib/components/global-search/command-trigger.svelte';
-	import { localizedHref } from '$lib/utils/i18n';
+	import { buildBreadcrumbs } from './breadcrumbs';
 	import { getTranslate } from '@tolgee/svelte';
 
 	const { t } = getTranslate();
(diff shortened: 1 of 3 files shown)


stickerdaniel:
* **#361** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/361?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/361?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #363, opened 2026-05-11
chore(deps): Add tanstack svelte table

kleenkanteen:
tanstack table is in `package.json` but is missing the svelte table package indicated here: https://tanstack.com/table/latest/docs/installation#svelte-table
image

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -23,6 +23,7 @@
         "@sentry/sveltekit": "^10.47.0",
         "@stickerdaniel/convex-autumn-svelte": "^0.2.0",
         "@svelte-put/lockscroll": "^2.0.1",
+        "@tanstack/svelte-table": "^8.21.3",
         "@tolgee/format-icu": "^7.0.0",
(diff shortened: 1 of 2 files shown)


stickerdaniel:
Thanks for the PR. This is actually intentional. The repo runs Svelte 5, and `@tanstack/svelte-table@8.21.3` peers `svelte: "^4.0.0 || ^3.49.0"` (Svelte 3/4 adapter).

For Svelte 5, the pattern is `@tanstack/table-core` plus an inline adapter, which is what `src/lib/components/ui/data-table/` already does. All 8 table imports in the codebase use `table-core` directly; nothing imports `@tanstack/svelte-table`.

The official shadcn-svelte data-table docs walk through the Svelte 5 setup if you're curious: https://shadcn-svelte.com/docs/components/data-table

Closing, but appreciate the look!

kleenkanteen:
makes sense, appreciate the explanation

PR #367, opened 2026-05-14
fix(env): default SITE_URL to localhost in convex example

### LINKED ISSUE #365: Local dev admin sign in does not work
Yesterday I used the default admin login and it worked, was using local convex instance. Booted it up today and I get a login error:

Image

output of `bun run dev`:

Image

I rolled back my changes and had the same issue. I looked at the convex endpoint and the frontend was calling port 5173:

Image

But convex is intialized at port 5968:

Image

ah calls to convex seem to be proxied?

here is another hint, brb:

Image

edit: found the fix in `.env.convex.local`, I had to change `SITE_URL=https://yourapp.com` to `SITE_URL=http://localhost:5173`:

Image

stickerdaniel:
Closes #365

Copying `.env.convex.example` to `.env.convex.local` left `SITE_URL=https://yourapp.com` as the local-dev default, which silently breaks Better Auth admin sign-in on the embedded backend (callback URL mismatch). The Convex-side env file is gitignored, so every fresh checkout reproduces the same trap.

Default `SITE_URL` in the example to `http://localhost:5173` to match `bun run dev`, and keep the cloud override discoverable via the existing comment (`bunx convex env set SITE_URL https://yourapp.com`). Added a one-liner under `AGENTS.md` "Local dev notes" so the constraint shows up where the rest of the local-backend gotchas live.

Static checks pass locally.

### DIFF
--- a/.env.convex.example
+++ b/.env.convex.example
@@ -76,8 +76,9 @@ AUTH_GITHUB_SECRET=
 # Base URL for OAuth redirects and deep links (can be preview URL)
 # Auto-derived from VERCEL_URL for preview deployments
 # Note: For email images, use EMAIL_ASSET_URL (always production)
-# Command: bunx convex env set SITE_URL https://yourapp.com
-SITE_URL=https://yourapp.com
+# Local dev default: http://localhost:5173 (matches `bun run dev`)
+# For cloud/prod, set via: bunx convex env set SITE_URL https://yourapp.com
+SITE_URL=http://localhost:5173
 
 # ====================================
 # E2E Testing (REQUIRED for test setup)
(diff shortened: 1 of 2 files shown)


stickerdaniel:
* **#367** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/367?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/367?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #368, opened 2026-05-14
feat(env): surface gated feature names in requireEnv errors

stickerdaniel:
First of a three-PR stack improving the dev experience around optional env keys. This bottom PR fixes the worst failure mode and lays the groundwork; PRs #369 and #370 build on top.

Without `RESEND_API_KEY` the Resend SDK accepts `sendEmail` calls and queues them, but the actual API call fails inside the component's workpool. Better Auth's verify-email flow then "succeeds" from the user's perspective while no email ever arrives, the verify link expires, and the only signal is component-internal log noise.

Extend `requireEnv` with an optional `{ feature, docs }` argument so the thrown error names the gated capability, prints the exact `bunx convex env set` fix command, and points at a reference doc. Every existing call site gets a feature label. Add an `assertResendApiKey` preflight to all six email handlers in `emails/send.ts` so unset Resend keys fail fast in Convex logs instead of vanishing into a workpool retry. The preflight no-ops when `AUTH_E2E_TEST_SECRET` is set so the E2E suite stays unaffected.

`bun run test:unit src/lib/convex/env.test.ts` passes (5 tests). `bun run check:convex` and `bun scripts/static-checks.ts` pass on every touched file.

### DIFF
--- a/src/lib/convex/auth.ts
+++ b/src/lib/convex/auth.ts
@@ -303,9 +303,9 @@ function buildTrustedOrigins(): string[] {
 // Creates Better Auth options object (used by adapter and betterAuth CLI)
 export const createAuthOptions = (ctx: GenericCtx<DataModel>): BetterAuthOptions => {
 	return {
-		baseURL: requireEnv('SITE_URL'),
+		baseURL: requireEnv('SITE_URL', { feature: 'Better Auth base URL' }),
 		trustedOrigins: buildTrustedOrigins(),
-		secret: requireEnv('BETTER_AUTH_SECRET'),
+		secret: requireEnv('BETTER_AUTH_SECRET', { feature: 'Better Auth session signing' }),
 		database: authComponent.adapter(ctx),
 		user: {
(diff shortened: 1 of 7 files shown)


stickerdaniel:
* **#370** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/370?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#369** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/369?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#368** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/368?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/368?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #369, opened 2026-05-14
feat(dev): emit console notices when optional features are gated

stickerdaniel:
Stacks on #368. Adds the in-context layer of the dev-experience improvement: when a developer hits a silently-degraded optional feature, log a one-time `[dev]` warning naming the feature and the env keys to set.

Today OAuth buttons disappear, PostHog/Sentry no-op, and Tolgee falls back to committed JSON without any signal to the developer about why. Resend webhook signature verification is the same shape — the component throws `"Webhook secret is not set"` from inside `handleResendEventWebhook` and Resend retries the 5xx response forever.

Introduce `src/lib/dev/notice.ts` plus a `DEV_FEATURES` catalog as the single source of truth for which env vars gate which capability. Wire the helper into every silent-degrade gate: `getAvailableOAuthProviders` for Google/GitHub sign-in, `initPosthog`, Sentry init on both client and server hooks, the Resend module-load for the webhook secret, and the Tolgee init in the root layout. Dedupe is per-process, so notices fire once per Convex isolate, server boot, or page session.

The helper is a no-op outside local dev (`LOCAL_CONVEX_DEV` for Convex, `import.meta.env.DEV` everywhere else), so production behavior is unchanged.

`bun run test:unit src/lib/dev/notice.test.ts` passes (6 tests). Full static checks pass on every touched file.

PR #375, opened 2026-05-16
feat(dev): emit console notices when optional features are gated

stickerdaniel:
Re-opened from #369 (auto-closed by GitHub when the stack-bottom PR #368 was merged and its base branch was deleted). Branch has been rebased onto current main; commits are identical to the original #369 fixes.

Today OAuth buttons disappear, PostHog/Sentry no-op, and Tolgee falls back to committed JSON without any signal to the developer about why. Resend webhook signature verification is the same shape — the component throws `"Webhook secret is not set"` from inside `handleResendEventWebhook` and Resend retries the 5xx response forever.

Introduce `src/lib/dev/notice.ts` plus a `DEV_FEATURES` catalog as the single source of truth for which env vars gate which capability. Wire the helper into every silent-degrade gate: `getAvailableOAuthProviders` for Google/GitHub sign-in, `initPosthog`, Sentry init on both client and server hooks, the Resend module-load for the webhook secret, and the Tolgee init in the root layout. Dedupe is per-process, so notices fire once per Convex isolate, server boot, or page session.

The helper is a no-op outside local dev (`LOCAL_CONVEX_DEV` for Convex, `import.meta.env.DEV` everywhere else), so production behavior is unchanged.

Builds on the merged #368.

### DIFF
--- a/src/hooks.client.ts
+++ b/src/hooks.client.ts
@@ -1,14 +1,23 @@
 import { dev, browser } from '$app/environment';
 import { PUBLIC_SENTRY_DSN } from '$env/static/public';
 import { getPosthog } from '$lib/analytics/posthog';
+import { devNotice } from '$lib/dev/notice';
 import * as Sentry from '@sentry/sveltekit';
 import type { HandleClientError } from '@sveltejs/kit';
 
-if (browser && PUBLIC_SENTRY_DSN) {
-	Sentry.init({
-		dsn: PUBLIC_SENTRY_DSN,
-		tracesSampleRate: 0.1
-	});
+if (browser) {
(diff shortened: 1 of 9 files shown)


stickerdaniel:
* **#375** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/375?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/375?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #393, opened 2026-05-20
Enable oxlint array-type rule + small UI polish

### LINKED ISSUE #374: Kbd hints in global-search command-trigger not vertically centered
## Problem

`src/lib/components/global-search/command-trigger.svelte` positions the trailing `Kbd.Group` with `absolute end-1.5 top-1.5` inside an `h-8` Button. The kbd row is `h-5.5` (22px), so:

- top: 6px
- bottom: 32 - 6 - 22 = 4px

The cluster sits 2px below geometric center. Combined with the placeholder text being centered by flexbox (cap-height-based), not by absolute `top:`, the `⌘ K` hints read as visibly offset next to "Search pages…".

## Reproduction

Render the global-search trigger on viewport >= sm. The `⌘ K` cluster sits low and a bit off relative to the placeholder text.

## Suggested fix

Replace the absolute `top-1.5` with `inset-y-0 items-center` so the group is symmetric to the Button height regardless of `Kbd.Root` height changes:

\`\`\`diff
-<div class="absolute end-1.5 top-1.5 hidden gap-1 sm:flex">
+<div class="absolute inset-y-0 end-1.5 hidden items-center gap-1 sm:flex">
\`\`\`

Happy to PR.

### LINKED ISSUE #380: Button is stale vs upstream vega: missing not-aria-[haspopup] guard on press effect
`src/lib/components/ui/button/button.svelte:7` has `active:translate-y-px`, which also fires on dropdown / popover triggers (anything with `aria-haspopup`). That makes the trigger jump 1px down on click while opening the menu, which feels glitchy.

Upstream vega has since shipped the smarter guard. Verify:

```
curl -s https://shadcn-svelte.com/registry/styles/vega/button.json | jq -r '.files[0].content' | grep -o 'active:[^ ]*'
# active:not-aria-[haspopup]:translate-y-px
```

Same pattern is present in luma, nova, maia, lyra, mira (verified for luma + vega).

This file was last touched in #244, where the override was reapplied on top of the older vega template. Upstream vega has moved on since.

Also missing from current button.svelte vs upstream vega:
- `aria-expanded:bg-muted` / `aria-expanded:bg-secondary` etc. on outline / secondary / ghost / destructive variants for dropdown-open feedback

### Fix

Option A (minimal): one-line swap on `base`:

```
- active:translate-y-px
+ active:not-aria-[haspopup]:translate-y-px
```

Option B (full re-sync): `bunx shadcn-svelte add button --overwrite`, then re-apply any deliberate project overrides on top.

Repro: any button used as a dropdown trigger (sidebar `nav-user`, table column menus, etc.). Click the trigger, watch the button drop 1px while the menu opens.

### LINKED ISSUE #383: Footgun: ESLint projectService fails on dirs outside tsconfig include (e.g. archive/, scratch/)
## Problem

When a fork or downstream project adds a new top-level directory like `archive/`, `scratch/`, `docs/snippets/` etc. and the directory contains `.svelte` / `.ts` files, pre-commit hooks fail with:

```
0:0  error  Parsing error: /path/to/archive/foo.svelte was not found by the project service. Consider either including it in the tsconfig.json or including it in allowDefaultProject
```

Root cause: the typescript-eslint `projectService: true` setting (in `eslint.config.js` Svelte block) refuses to parse any file not covered by the resolved tsconfig.

The error is correct in spirit (file is genuinely outside the type-checked project), but the failure mode is opaque for a fork maintainer who just dropped reference snippets in `archive/` and didn't expect a typed-lint pass on them.

## Repro

In a fresh saas-starter fork:

```bash
mkdir archive && cat > archive/page.svelte <<'SVELTE'
<script lang="ts">
  let x = 1;
</script>
<p>{x}</p>
SVELTE
git add archive/
git commit -m "test"
# → hook fails at the eslint stage with the parser error above
```

## Suggested fixes (any one helps)

1. **Defensive default ignores** in `eslint.config.js:30` for common fork-only dirs:
   ```js
   ignores: [
     '**/_generated/**',
     'src/env.d.ts',
     'src/lib/convex/convex-env.d.ts',
     // Footgun protection for forks: typed-lint refuses files outside tsconfig project
     'archive/**',
     'scratch/**',
     '.snapshots/**'
   ]
   ```
2. **AGENTS.md / docs note** under "Cross-Platform Scripts" or a new "ESLint footguns" section:
   > Adding a new top-level dir with `.svelte`/`.ts` files outside the tsconfig include? You'll hit a `projectService` parsing error in pre-commit. Either add the dir to `tsconfig.json` `include`, or add it to the top-level `ignores` block in `eslint.config.js`.
3. **Pre-commit warning** in `scripts/static-checks.ts` that detects staged files outside the resolved tsconfig and prints a helpful pointer before the cryptic ESLint error surfaces.

Happy to PR option 1 + 2 if desired.

## Context

Hit this in a downstream fork that uses `archive/` as a holding spot for old route files kept as design references. The fix (one line in `eslint.config.js` ignores) is trivial once you know — but the error message is misleading enough that I initially misdiagnosed it as a `scripts/static-checks.ts` bug handling deleted files (it isn't, that path correctly uses `--diff-filter=ACMR`).

### LINKED ISSUE #392: Style drift: Array<T> vs T[] in support/messageListing.ts and support/ownership.ts
The rest of the repo consistently uses the postfix \`T[]\` notation for array types (e.g. \`ChatMessage[]\`, \`string[]\`), but two support-module files mix in the \`Array<T>\` generic form:

\`src/lib/convex/support/messageListing.ts\` (6 spots):
- line 49: \`{ page: Array<RawMessageRow> }\`
- line 123: \`page: Array<ChatMessage>\`
- line 126: \`Promise<Array<ChatMessage>>\`
- line 146: \`page: Array<ChatMessage>\`
- line 147: \`materializedStreams: Array<UIMessage>\`
- line 148: \`): Array<ChatMessage>\`

\`src/lib/convex/support/ownership.ts\` (1 spot):
- line 108: \`as Array<AgentMessageLookupResult | null>\`

The \`cursors: Array<{ streamId: string; cursor: number }>\` at line 24 of \`messageListing.ts\` is legitimately more readable as \`Array<...>\` because the alternative \`({...})[]\` requires extra parens.

Suggested fix: switch the seven flagged spots to postfix \`T[]\` for consistency with the rest of the codebase.

Purely cosmetic, no behavior change.

stickerdaniel:
Closes #392
Closes #374
Closes #380
Closes #383

Four small fixes from the issue tracker, each as its own commit so they can be reviewed independently.

`chore(lint): enforce postfix T[] array notation via oxlint` activates the `typescript/array-type` rule with the `array-simple` option in `.oxlintrc.json` and applies the auto-fix across the repo. The simple-vs-complex split means inline-object types like `Array<{ streamId: string; cursor: number }>` are still allowed where they read better.

`fix(ui): center kbd hints in global-search trigger` swaps `top-1.5` for `inset-y-0 items-center` on the absolute Kbd wrapper so the ⌘K cluster stays symmetric to the trigger button height.

`fix(ui): skip press effect on dropdown triggers` guards the active translate with `not-aria-[haspopup]` so dropdown/popover triggers no longer drop 1px while opening the menu. Brings the button base in line with current upstream vega.

`docs(agents): note projectService limitation for dirs outside tsconfig` adds a one-line note under ESLint & Legacy Plugins explaining the cryptic projectService parsing error when a fork adds a top-level dir outside `tsconfig.json` `include`. No code changes - the configuration footgun is intentional behaviour of `projectService: true`, and a defensive ignore list of magic names (`archive/`, `scratch/`) would only cover the issue reporter's specific naming.

`bun scripts/static-checks.ts --staged` passes for each commit. No behaviour change beyond the documented fixes.

### DIFF
--- a/.oxlintrc.json
+++ b/.oxlintrc.json
@@ -10,6 +10,7 @@
 				"ignoreUsageFiles": ["**/*.test.ts", "e2e/**"]
 			}
 		],
-		"convex/no-invalid-module-path": "error"
+		"convex/no-invalid-module-path": "error",
+		"typescript/array-type": ["error", { "default": "array-simple" }]
 	}
 }
--- a/.oxlintrc.json
+++ b/.oxlintrc.json
@@ -11,6 +11,7 @@
 			}
 		],
 		"convex/no-invalid-module-path": "error",
-		"typescript/array-type": ["error", { "default": "array-simple" }]
+		"typescript/array-type": ["error", { "default": "array-simple" }],
+		"typescript/consistent-type-imports": ["error", { "disallowTypeAnnotations": true }]
 	}
 }
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -397,6 +397,8 @@ Prop names must match the parent's passed prop name exactly.
 
 When adding ESLint plugins that export legacy `.eslintrc`-style configs (objects with `overrides`), use `fixupConfigRules()` from `@eslint/compat` to convert. See the Convex plugin block in `eslint.config.js` for the pattern.
 
+Adding a top-level dir with `.svelte`/`.ts` files outside the tsconfig `include` (e.g. `archive/`, `scratch/`)? `projectService: true` will refuse to parse it and pre-commit fails with a cryptic "not found by the project service" error. Either add the dir to `tsconfig.json` `include` or to `eslint.config.js` `ignores`.
+
 ### Real-time Features
 
 - Use Convex's `useQuery` for reactive data — **never inside `$derived()`**, use the `'skip'` pattern instead
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -77,7 +77,11 @@ export default defineConfig(
 			'@typescript-eslint/ban-ts-comment': 'error',
 
 			// Off globally: ~26 legitimate usages in UI-kit/vendor code (shadcn, Konva, Rive, TanStack)
-			'@typescript-eslint/no-explicit-any': 'off'
+			'@typescript-eslint/no-explicit-any': 'off',
+
+			// Mirrors oxlint typescript/consistent-type-imports — needed here because oxlint
+			// does not yet parse <script lang="ts"> blocks inside .svelte files
+			'@typescript-eslint/consistent-type-imports': ['error', { disallowTypeAnnotations: true }]
 		}
 	},
 	// Convex-specific: suppress type-checked rules that don't apply to Convex handler patterns
--- a/eslint/rules/no-hardcoded-aria-label.test.ts
+++ b/eslint/rules/no-hardcoded-aria-label.test.ts
@@ -2,8 +2,8 @@ import { describe, expect, it } from 'vitest';
 import parser from 'svelte-eslint-parser';
 import rule from './no-hardcoded-aria-label.js';
 
-function lint(template: string): { messageId: string }[] {
-	const reports: { messageId: string }[] = [];
+function lint(template: string): Array<{ messageId: string }> {
+	const reports: Array<{ messageId: string }> = [];
 	const ast = parser.parseForESLint(template, {});
 
 	const context = {
--- a/eslint/rules/no-hardcoded-sr-only.test.ts
+++ b/eslint/rules/no-hardcoded-sr-only.test.ts
@@ -2,8 +2,8 @@ import { describe, expect, it } from 'vitest';
 import parser from 'svelte-eslint-parser';
 import rule from './no-hardcoded-sr-only.js';
 
-function lint(template: string): { messageId: string }[] {
-	const reports: { messageId: string }[] = [];
+function lint(template: string): Array<{ messageId: string }> {
+	const reports: Array<{ messageId: string }> = [];
 	const ast = parser.parseForESLint(template, {});
 
 	const context = {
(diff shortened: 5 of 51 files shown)


stickerdaniel:
* **#393** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/393?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/393?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

stickerdaniel:
Triage of Copilot's 6 array-type findings:

**Valid (2)** — fixed in eac7089a:
- `authenticated-sidebar.svelte:21` and `app/+layout.svelte:98` used inline `import('./types').NavSubItem` which oxlint treats as non-simple. Promoted to a named top-level import so the postfix `NavSubItem[]` notation matches the other five call sites.

**False positive (4)** — `columns.ts` x2, `data-table.svelte.ts:91`, `notifications.ts:481`:
- `ColumnDef<AdminUserData>`, `ColumnDef<NotificationRecipient>`, `MaybeThunk<any>`, `Id<'messages'>` are all generic types with type arguments. Empirically verified oxlint classifies these as non-simple under `array-simple` and rewrites postfix `T[]` back to `Array<T>`. Reverting to `ColumnDef<X>[]` per Copilot's suggestion would re-fail the lint rule this PR introduces.

PR #394, opened 2026-05-20
Fix support widget z-index, PersistedState supportUserId, deploy alias slice, RouteProgress timer leak

### LINKED ISSUE #373: z-index of support widget is higher than the preview modal
Image

### LINKED ISSUE #376: sanitizeBranchAlias slice limit is hardcoded for old slug length
## Problem

`scripts/deploy/platform.ts` truncates the sanitized branch alias to a hardcoded 50 chars so the CF Workers preview hostname stays within the 63-char DNS label limit:

```ts
// scripts/deploy/platform.ts:22
.slice(0, 50) // DNS-aware: 63 - 1 (dash) - 12 (saas-starter) = 50
```

The math (`63 - 1 - 12`) is baked against `saas-starter` (12 chars). The preview hostname is `${alias}-${workerName}.${subdomain}.workers.dev`, and `workerName` comes from `wrangler.toml` `name =`, which `bun run setup` rewrites. So after a rebrand:

- A shorter worker name (e.g. `myapp`, 5 chars) wastes 7 chars of branch slug — branches get truncated more aggressively than necessary.
- A longer worker name (e.g. `my-cool-saas-product`, 20 chars) produces hostnames that exceed the 63-char DNS label limit on long branch names, which CF rejects.

Neither failure is loud — the first silently truncates branch names, the second only surfaces when a long-named branch is pushed.

## Suggested fix

Compute the slice limit from the actual worker name length. Either:

**Option A — pass worker name as parameter:**

```ts
export function sanitizeBranchAlias(branch: string, workerName: string): string {
	const maxAliasLen = 63 - 1 - workerName.length;
	return branch
		.toLowerCase()
		...
		.slice(0, maxAliasLen)
		.replace(/-$/, '') || 'branch';
}
```

Callers (`detectPlatform` at line 68, `cf-deploy.ts` at line 20) already have `workerName` (from `process.env.WORKERS_NAME` / `wrangler.toml`) in scope, so plumbing it through is trivial.

**Option B — read worker name from `wrangler.toml`** inside the function. More coupling, harder to test.

I'd lean toward Option A. Happy to open a PR.

## Related

This shows up as part of the [non-TTY `bun run setup` issue (#372)](https://github.com/stickerdaniel/saas-starter/issues/372) — the comment in the file even references the old slug length explicitly, which is a tell that this was meant to be dynamic but never made it.

### LINKED ISSUE #384: RouteProgress: timer leak on mid-navigation unmount (missing onDestroy cleanup)
## Problem

`src/lib/components/RouteProgress.svelte` schedules four cascade `setTimeout` / `setInterval` timers (`startDelayTimer`, `trickleTimer`, `completeTimer`, `fadeTimer`) inside `$effect` reactions to `navigating`, but it does **not** clean them up on component unmount. If the component is destroyed mid-navigation (e.g. user logs out, sub-tree re-mounts, top-level layout swap), the timers keep firing against detached state until they self-resolve.

The `clearTimers()` helper exists but is only called from `startProgress()` / `completeProgress()` entry points. There is no `onDestroy(clearTimers)`.

## Repro

1. Trigger navigation that takes >100ms (so `startDelayTimer` fires).
2. Before navigation completes, unmount the layout containing `<RouteProgress />` (e.g. force a full reload, switch auth state).
3. Observe in DevTools that `trickleTimer` continues writing to detached `progress` state.

## Suggested fix

Two-line cleanup + a runed simplification of the three setTimeouts:

```svelte
<script lang="ts">
  import { useDebounce } from 'runed';
  import { onDestroy } from 'svelte';

  // Replace startDelayTimer/completeTimer/fadeTimer with cancellable useDebounce
  // calls; keep trickleTimer as setInterval since it needs the recurring tick.
  const scheduleShow = useDebounce(showProgress, START_DELAY_MS);
  const scheduleFadeStart = useDebounce(() => {
    opacity = 0;
    scheduleFadeFinish();
  }, SPEED);
  const scheduleFadeFinish = useDebounce(() => {
    showBar = false; opacity = 1; progress = 0;
  }, SPEED);

  function clearTimers() {
    scheduleShow.cancel();
    scheduleFadeStart.cancel();
    scheduleFadeFinish.cancel();
    if (trickleTimer) { clearInterval(trickleTimer); trickleTimer = null; }
  }
  onDestroy(clearTimers);
</script>
```

Eliminates 3 local timer variables, gets auto-cleanup of the debounce timers on unmount, and the explicit `onDestroy(clearTimers)` handles the interval.

Working diff verified in a downstream fork; happy to send a PR if useful.

### LINKED ISSUE #387: Customer-support widget: raw localStorage for supportUserId should be PersistedState
## Problem

`src/lib/components/customer-support/customer-support.svelte:83-86` reads + writes `supportUserId` directly through `localStorage` API:

\`\`\`ts
let id = localStorage.getItem('supportUserId');
if (!id) {
  id = generateAnonymousUserId();
  localStorage.setItem('supportUserId', id);
}
\`\`\`

The rest of the codebase uses runed's \`PersistedState\` for any cross-session value (last-auth-method, ChatDraftManager, mode-watcher theme, etc.). This single hold-out is inconsistent and loses the reactive + cross-tab sync behaviour PersistedState provides.

## Why this matters

- No reactive read — components reading the id can't subscribe to changes
- No cross-tab sync (if the user opens a second tab the support-thread state diverges)
- The literal key 'supportUserId' is anchor for future migrations; harder to grep than a centralised constant

## Suggested fix

\`\`\`ts
import { PersistedState } from 'runed';
const supportUserId = new PersistedState<string | null>('supportUserId', null);
// Usage:
function getAnonymousUserId() {
  if (!supportUserId.current) supportUserId.current = generateAnonymousUserId();
  return supportUserId.current;
}
\`\`\`

## See also

- #384, #385, #386 — same runed-consistency theme

stickerdaniel:
Closes #373
Closes #387
Closes #376
Closes #384

Four small fixes bundled as separate commits.

The customer-support FAB and AI chatbar rendered above shadcn dialog and sheet overlays including the chat attachment file-preview dialog, so any modal triggered from inside the support flow ended up partially hidden behind it. Lowered the three FAB/chatbar containers from z-200 / z-[100] to z-40 so all standard modals (z-50) and the screenshot editor (z-[100]) sit above the widget.

Replaced the two raw localStorage call sites for the anonymous supportUserId with a shared PersistedState helper so the value goes through one source of truth. The helper uses a custom serializer with a JSON-parse-then-raw-string fallback to preserve backward compatibility with values written by previous releases, and the initial value is undefined so PersistedState does not eagerly write to storage on import.

The CF Workers preview-alias slice limit was hardcoded to 50 chars based on the original "saas-starter" worker name. After a rebrand a shorter worker name wasted alias budget and a longer worker name could produce preview hostnames exceeding the 63-char DNS label limit silently. Pass workerName as a required parameter to sanitizeBranchAlias and compute the slice budget dynamically. Throw early when the remaining budget is below the "branch" fallback length to avoid the negative-slice trap where 'foo'.slice(0, -1) returns 'fo'.

RouteProgress.svelte started four cascade setTimeout/setInterval timers from $effect reactions but never cleaned them up on destroy. If the layout unmounted mid-navigation (logout, full reload, layout swap) the timers kept writing to detached reactive state. Added onDestroy(clearTimers).

cd saas-starter && bun scripts/static-checks.ts on each changed file passes. bun run test:unit -- platform.test.ts adds 8 new boundary-case tests covering short/long worker names, digit-leading branches, empty/all-punctuation branches, and the exact-budget and over-budget boundaries.

### DIFF
--- a/scripts/cf-deploy.ts
+++ b/scripts/cf-deploy.ts
@@ -17,7 +17,12 @@ const isPreview = branch !== undefined && branch !== productionBranch;
 const args = ['wrangler', 'versions', 'upload'];
 
 if (isPreview && branch) {
-	const alias = sanitizeBranchAlias(branch);
+	const workerName = process.env.WORKERS_NAME;
+	if (!workerName) {
+		console.error('WORKERS_NAME is required to compute the preview alias slice limit.');
+		process.exit(1);
+	}
+	const alias = sanitizeBranchAlias(branch, workerName);
 	args.push('--preview-alias', alias);
 	console.log(`Preview alias: ${alias}`);
 }
--- a/scripts/deploy/platform.test.ts
+++ b/scripts/deploy/platform.test.ts
@@ -1,0 +1,54 @@
+import { describe, expect, it } from 'vitest';
+import { sanitizeBranchAlias } from './platform';
+
+describe('sanitizeBranchAlias', () => {
+	it('uses the full alias budget for short worker names', () => {
+		const workerName = 'myapp'; // 5 chars, budget 57
+		const longBranch = 'a'.repeat(80);
+		const alias = sanitizeBranchAlias(longBranch, workerName);
+		expect(alias.length).toBe(57);
+		expect(`${alias}-${workerName}`.length).toBeLessThanOrEqual(63);
+	});
+
+	it('shrinks alias budget for long worker names', () => {
+		const workerName = 'my-cool-saas-product'; // 20 chars, budget 42
+		const longBranch = 'a'.repeat(80);
+		const alias = sanitizeBranchAlias(longBranch, workerName);
+		expect(alias.length).toBe(42);
+		expect(`${alias}-${workerName}`.length).toBeLessThanOrEqual(63);
+	});
+
+	it('keeps digit-leading branches valid even with long worker names', () => {
+		const workerName = 'my-cool-saas-product'; // 20 chars, budget 42
+		const alias = sanitizeBranchAlias('123-feature-branch', workerName);
+		expect(alias.startsWith('b-1')).toBe(true);
+		expect(alias).toMatch(/^[a-z]/);
+		expect(`${alias}-${workerName}`.length).toBeLessThanOrEqual(63);
+	});
+
+	it('falls back to "branch" for an empty branch name', () => {
+		expect(sanitizeBranchAlias('', 'myapp')).toBe('branch');
+	});
+
+	it('falls back to "branch" for an all-punctuation branch name', () => {
+		expect(sanitizeBranchAlias('---', 'myapp')).toBe('branch');
+	});
+
+	it('throws when the worker name leaves a budget below the fallback length', () => {
+		const tooLong = 'a'.repeat(57); // budget = 5, fallback "branch" is 6
+		expect(() => sanitizeBranchAlias('feature', tooLong)).toThrow(/leaves only 5 chars/);
+	});
+
+	it('accepts a worker name that leaves a budget exactly equal to the fallback length', () => {
+		const justFits = 'a'.repeat(56); // budget = 6, exactly fits "branch"
+		expect(sanitizeBranchAlias('', justFits)).toBe('branch');
+	});
+
+	it('trims trailing dash introduced by truncation', () => {
+		const workerName = 'myapp';
+		// 56-char branch ending with chars that will keep a dash at slice boundary
+		const branch = 'a'.repeat(56) + '-tail';
+		const alias = sanitizeBranchAlias(branch, workerName);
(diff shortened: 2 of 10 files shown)


stickerdaniel:
* **#394** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/394?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/394?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #395, opened 2026-05-21
fix(setup): accept CLI flags, fail loud on non-TTY

### LINKED ISSUE #372: bun run setup hangs on third prompt with non-TTY stdin
## Problem

`scripts/template-setup.ts` uses `readline.createInterface` with `rl.question()` for three sequential prompts (slug, repo, brand). When stdin is not a TTY (piped input, redirected file, CI, or any non-interactive invocation), readline closes after consuming the buffered lines and the third `rl.question()` call never resolves. The script exits cleanly with code 0 having written nothing.

## Reproduction

```bash
printf "my-app\nowner/my-app\nMy App\n" | bun run setup
```

Expected: all three prompts answered, files patched.
Actual: only the first two prompts print, the script exits 0, no files are modified.

Same behaviour with `node`/`tsx` and with `here-doc` input. Adding extra trailing newlines or `sleep` after the printf does not help — the readline interface is closed before the third `question()` callback can register.

## Why it matters

- Blocks any automation that wants to bootstrap a fresh template (CI provisioning, scripted scaffolding, AI agents driving the setup).
- Silent failure (exit 0, no error) makes the bug hard to diagnose.

## Suggested fix

Two changes to `scripts/template-setup.ts`:

1. Accept the three values as CLI flags (`--slug`, `--repo`, `--brand`), falling back to the prompt when the flag is missing.
2. When stdin is not a TTY and a flag is missing, use the detected current value rather than blocking on `rl.question()`.

Result:

```bash
bun run setup --slug my-app --repo owner/my-app --brand "My App"
```

works end-to-end without needing a real terminal. Interactive `bun run setup` still works exactly as before for users following the README.

I have a working patch applied locally — happy to open a PR if useful.

stickerdaniel:
Closes #372

`bun run setup` hangs when stdin is not a TTY (piped input, CI, scripted scaffolding, AI agents). `scripts/template-setup.ts` constructs `readline.createInterface` and runs three sequential `rl.question()` calls for slug, repo, and brand. Readline closes after consuming the buffered lines from non-TTY stdin and the third callback never fires, so the awaited promise never resolves. The script exits cleanly with code 0 and writes nothing. Silent failure, hard to diagnose.

This change adds `--slug`, `--repo`, `--brand`, and `--help` via `node:util` `parseArgs`, mirroring the existing pattern in `scripts/create-worktree.ts`. No new dependency. Interactive runs are unchanged: missing flags still prompt with current values as defaults. Non-interactive runs now require all three flags; missing values produce a single error listing exactly which flags are missing plus a copy-paste example, then exit 1. The fallback-to-current-values path was rejected because it would let a misconfigured CI run silently write the unchanged template values back into `package.json`, `README.md`, etc., shifting the silent-failure class rather than removing it.

Readline is now constructed lazily and only when an interactive prompt is actually needed, so pure flag-driven runs skip it entirely.

Verified: `printf "..." | bun run setup` now fails loud with exit 1. `bun run setup --slug my-test --repo owner/my-test --brand "My Test"` patches all nine template files end-to-end. Partial-flag interactive runs prompt only for the missing flags. `--help` prints usage and exits 0. Bad slug/repo input from flags still trips the existing validators. `bun scripts/static-checks.ts scripts/template-setup.ts` passes.

### DIFF
--- a/scripts/template-setup.ts
+++ b/scripts/template-setup.ts
@@ -4,15 +4,56 @@
  *
  * Safe to re-run: prompts with current values as defaults.
  *
- * Usage: bun run setup
+ * Usage:
+ *   bun run setup
+ *   bun run setup --slug my-app --repo owner/my-app --brand "My App"
+ *
+ * Non-interactive mode (piped stdin, CI, agents) requires all three flags.
  */
 
 import { readFileSync, writeFileSync } from 'fs';
 import { join } from 'path';
-import { createInterface } from 'readline';
+import { createInterface, type Interface } from 'readline';
+import { parseArgs } from 'util';
 
 const ROOT = join(import.meta.dirname, '..');
-const rl = createInterface({ input: process.stdin, output: process.stdout });
+
+const { values } = parseArgs({
+	args: Bun.argv.slice(2),
+	options: {
+		slug: { type: 'string' },
+		repo: { type: 'string' },
+		brand: { type: 'string' },
+		help: { type: 'boolean', short: 'h', default: false }
+	},
+	strict: false,
+	allowPositionals: false
+});
+
+if (values.help) {
+	console.log(`Template Setup
+
+Usage:
+  bun run setup                                          (interactive)
+  bun run setup --slug <s> --repo <owner/name> --brand <s>
+
+Flags:
+  --slug    Project slug (lowercase, hyphens; matches ^[a-z0-9-]+$)
+  --repo    GitHub repo in owner/name format
+  --brand   Brand display name
+  -h, --help  Show this help
+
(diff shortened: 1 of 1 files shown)


stickerdaniel:
* **#395** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/395?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/395?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #401, opened 2026-05-23
fix(error-page): Add clickable bug-report mailto link

### LINKED ISSUE #398: Add clickable bug-report mailto link on generic error page
The current `ErrorDisplay.svelte` shows a generic "An unexpected error occurred while loading this page." with no contact path. Users hitting a 500 have no obvious next step beyond going home, and any bug report they do send to the contact address arrives without URL or status code, so the operator has to play 20 questions to reproduce.

Polish: on non-404 errors, render a clickable contact email built from `LEGAL_CONFIG.email` that opens a `mailto:` with a pre-filled, localized bug report template containing the failed URL and HTTP status.

Suggested shape (sketch, not a PR):

- `src/i18n/<lang>.json` — add `error_page.bug_report_subject` and `error_page.bug_report_body` keys with `{url}` and `{status}` placeholders, and switch `generic_description` from a literal email to a `{email}` placeholder so `LEGAL_CONFIG.email` stays the single source of truth.
- `src/lib/components/ErrorDisplay.svelte` —
  - import `getLegalEmailAddress` from `$lib/config/legal`,
  - build a `$derived` `mailtoHref` that runs `.replace('{url}', …).replace('{status}', …)` on the subject + body and `encodeURIComponent`s the result,
  - split the description on `{email}` (as a `$derived` so locale and status changes recompute) and render `prefix + <a href={mailtoHref}>{email}</a> + suffix`.
- Keep the link visually as inline text — same `text-muted-foreground` as surrounding description, underline from `Empty.Description`'s default `[&>a]:underline`, but override the `hover:text-primary` with `hover:!text-current` so the link does not look like a primary CTA inside an error empty-state.

Why this fits the template:
- Uses `LEGAL_CONFIG.email` that already exists.
- 404 path is unchanged (no email, no bug report — by design).
- Stays inside the existing `Empty.*` + i18n patterns; no new component, no new dependency.
- Auto-attaching URL + status removes the most common "what page were you on / what was the error code" follow-up email roundtrip.

## Implementation gotchas

Two things surfaced while shipping this in a fork. Bake them into the implementation or the click silently does nothing:

**1. Do not click-navigate the mailto via the href alone.** A plain `<a href={mailtoHref}>` gets intercepted by SvelteKit's client router and the click silently does nothing. Adding `target="_blank"` is worse: it spawns an empty `about:blank` tab and tries to navigate it to the mailto. Keep the `href` (for right-click "Copy link" and a11y) but trigger the open from an `onclick` handler:

```svelte
<a href={mailtoHref} onclick={openMailto}>{email}</a>
```

```ts
function openMailto(event: MouseEvent): void {
  event.preventDefault();
  window.location.href = mailtoHref;
}
```

This sets the location on the current page exactly like typing the mailto in the address bar, which reliably hands off to the OS mail handler.

**2. The body needs CRLF line breaks, not bare LF.** Apple Mail and other RFC-6068-strict clients silently refuse to open a mailto whose body uses `\n` (encoded `%0A`). They need `\r\n` (`%0D%0A`). Since the body template lives in i18n JSON with `\n`, normalize before encoding:

```ts
const body = encodeURIComponent(
  template.replace('{url}', url).replace('{status}', status).replace(/\r?\n/g, '\r\n')
);
```

Symptom without this: the link looks correct, the encoded href is correct, but the click opens nothing. The encoding itself was never the problem (browser status bars decode `%20` for display, which is a red herring while debugging this).

Happy to send a PR if this direction makes sense.

stickerdaniel:
Closes #398

`ErrorDisplay.svelte` showed a generic "An unexpected error occurred while loading this page." with no contact path. Users hitting a 500 had no obvious next step beyond going home, and any bug report they did send arrived with no URL or status code, so reproducing it required a back-and-forth.

On non-404 errors, `generic_description` now includes an `{email}` placeholder that resolves to `LEGAL_CONFIG.email` and renders as a clickable inline `mailto:` link. The mailto opens with a localized subject and body pre-filled with the failed URL and HTTP status. The link inherits the surrounding `text-muted-foreground` color (with the existing `Empty.Description` underline) and overrides the parent's `hover:text-primary` with `hover:!text-current`, so it reads as inline text rather than a primary CTA inside the error empty-state. 404 path is unchanged.

The href is built by a pure `buildBugReportMailto` helper (`src/lib/utils/mailto.ts`) that normalizes the i18n body template's line breaks to CRLF before percent-encoding, because Apple Mail and other RFC-6068-strict clients silently refuse to open a mailto whose body uses bare LF (`%0A`). Clicking assigns `window.location.href` directly so the OS mail handler opens in the current tab; `target="_blank"` spawned an empty `about:blank` tab. The plain href stays on the anchor for right-click copy and accessibility.

Translations added in all four locales (`en`, `de`, `es`, `fr`): `bug_report_subject` with `{status}` and `bug_report_body` with `{url}` + `{status}`. `bun scripts/static-checks.ts` and `bun run test:unit` pass; new unit tests cover CRLF encoding and `{url}`/`{status}` substitution. Visually verified at `http://localhost:5173/en/<unknown-path>` that the email link matches the muted body text color (`oklch(0.705 0.015 286.067)`), not the primary `Back to home` color. The actual handoff to a desktop mail client was not re-verified after the CRLF change.

### DIFF
--- a/src/i18n/de.json
+++ b/src/i18n/de.json
@@ -1665,7 +1665,9 @@
 	},
 	"error_page": {
 		"back_home": "Zurück zur Startseite",
-		"generic_description": "Beim Laden dieser Seite ist ein unerwarteter Fehler aufgetreten.",
+		"bug_report_body": "Hallo,\n\nbeim Verwenden von SaaS Starter ist ein Fehler aufgetreten.\n\nURL: {url}\nStatus: {status}\n\nWas ich gerade gemacht habe:\n\n\nDanke!",
+		"bug_report_subject": "Fehlerbericht: Fehler {status} in SaaS Starter",
+		"generic_description": "Beim Laden dieser Seite ist ein unerwarteter Fehler aufgetreten. Falls das Problem weiterhin besteht, melden Sie es bitte an {email}.",
 		"generic_title": "Etwas ist schiefgelaufen",
 		"need_help": "Brauchen Sie Hilfe?",
 		"not_found_description": "Die gesuchte Seite existiert nicht. Versuchen Sie unten nach dem zu suchen, was Sie brauchen.",
--- a/src/i18n/en.json
+++ b/src/i18n/en.json
@@ -1666,7 +1666,9 @@
 	},
 	"error_page": {
 		"back_home": "Take me home",
-		"generic_description": "An unexpected error occurred while loading this page.",
+		"bug_report_body": "Hi,\n\nI hit an error while using SaaS Starter.\n\nURL: {url}\nStatus: {status}\n\nWhat I was doing:\n\n\nThanks!",
+		"bug_report_subject": "Bug report: error {status} on SaaS Starter",
+		"generic_description": "An unexpected error occurred while loading this page. If this keeps happening, please report it to {email}.",
 		"generic_title": "Something went wrong",
 		"need_help": "Need help?",
 		"not_found_description": "The page you're looking for doesn't exist. Try searching for what you need below.",
--- a/src/i18n/es.json
+++ b/src/i18n/es.json
@@ -1665,7 +1665,9 @@
 	},
 	"error_page": {
 		"back_home": "Volver al inicio",
-		"generic_description": "Ocurrió un error inesperado al cargar esta página.",
+		"bug_report_body": "Hola,\n\nEncontré un error mientras usaba SaaS Starter.\n\nURL: {url}\nEstado: {status}\n\nLo que estaba haciendo:\n\n\n¡Gracias!",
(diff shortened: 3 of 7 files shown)


stickerdaniel:
* **#401** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/401?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/401?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #403, opened 2026-05-23
fix(dev): avoid dynamic import.meta.env access in dev notice helper

stickerdaniel:
SvelteKit dev was crashing every request with HTTP 500 because Vite's module-runner now rejects dynamic property access on `import.meta.env` ("Dynamic access of import.meta.env is not supported"). The crash happened in `src/lib/dev/notice.ts` via `hooks.server.ts:21`, before any `+error.svelte` could render, so every page on `bun run dev` looked broken.

The notice helper had to dodge `import.meta.env.DEV` originally because the module is imported from both the Vite tsconfig (SvelteKit, browser) and the Convex tsconfig (Convex isolates), and the Convex tsconfig has no Vite types. The previous workaround was a cast to a partial type, but the runtime access was still dynamic, which is what Vite's module-runner objects to.

Switched the SvelteKit/browser dev check from `import.meta.env.DEV` to `process.env.NODE_ENV === 'development'`. SvelteKit SSR gets NODE_ENV from Node at runtime, and Vite statically replaces `process.env.NODE_ENV` in the browser bundle, so the check works in both environments. `@types/node` covers it under both tsconfigs, no `@ts-expect-error` needed. The Convex-scope branch is unchanged.

The Vitest case now sets `NODE_ENV='development'` in `beforeEach` and restores it in `afterEach` so the live "logs a vite-public feature" assertion still exercises the dev path.

`bun scripts/static-checks.ts` and `bun run test:unit src/lib/dev/notice.test.ts` pass. Verified live: `curl http://localhost:5175/en` returned HTTP 500 before the fix and HTTP 200 after, on a fresh dev server in this worktree.

### DIFF
--- a/src/lib/dev/notice.test.ts
+++ b/src/lib/dev/notice.test.ts
@@ -3,10 +3,12 @@ import { _resetDevNoticeDedupeForTests, devNotice } from './notice';
 
 describe('devNotice', () => {
 	const originalLocalConvexDev = process.env.LOCAL_CONVEX_DEV;
+	const originalNodeEnv = process.env.NODE_ENV;
 	let warnSpy: ReturnType<typeof vi.spyOn>;
 
 	beforeEach(() => {
 		_resetDevNoticeDedupeForTests();
+		process.env.NODE_ENV = 'development';
 		warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
 	});
 
@@ -16,6 +18,11 @@ describe('devNotice', () => {
 			delete (process.env as Record<string, string | undefined>).LOCAL_CONVEX_DEV;
 		} else {
 			process.env.LOCAL_CONVEX_DEV = originalLocalConvexDev;
+		}
+		if (originalNodeEnv === undefined) {
+			delete (process.env as Record<string, string | undefined>).NODE_ENV;
+		} else {
+			process.env.NODE_ENV = originalNodeEnv;
 		}
 	});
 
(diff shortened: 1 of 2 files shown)


stickerdaniel:
* **#403** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/403?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/403?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

stickerdaniel:
Closing. The same dev-SSR issue was already fixed on main in #414, and after looking more carefully I think the `import.meta.env.DEV` route (the one merged) is the correct one here, not the `process.env.NODE_ENV` route this PR takes.

Reason: `devNotice` is called from `src/hooks.client.ts` and `src/lib/analytics/posthog.ts`, so it runs in the client bundle too. Vite replaces the literal `process.env.NODE_ENV` there, but it does not shim the `process` global, so the required `typeof process !== 'undefined'` guard returns false at runtime in the browser. Result: `isDev()` would always be false in the browser, and every browser-side `devNotice` call would silently no-op.

`import.meta.env.DEV` is the only form Vite statically replaces in both client and SSR builds, which is why the merged version keeps it (with the inline cast for the Convex tsconfig). I just pushed a follow-up commit to `notice.ts` documenting this, so the next cleanup attempt does not retry the NODE_ENV approach.

The new test cases you added in `notice.test.ts` are nice. If you want, I can port them as a separate PR adapted to the import.meta.env.DEV mechanism (Vitest already sets `import.meta.env.DEV === true` by default, so the test setup is even simpler).

PR #402, opened 2026-05-23
refactor(support): replace setTimeout with useDebounce in widgets

stickerdaniel:
Closes #386

Two customer-support widgets kept ad-hoc `setTimeout` + `clearTimeout` plumbing for cancellable one-shot timers: a fallback that force-resolves the avatar-load animation after `ANIMATION_TIMEOUT` ms in `threads-overview.svelte`, and a self-clearing rate-limit timer in `feedback-widget.svelte`. Both patterns are exactly what runed's `useDebounce` covers, with the bonus that the wait can be a reactive getter (used in `feedback-widget` to compute `rateLimitedUntil - Date.now()` lazily on each effect re-run).

The one gotcha: `useDebounce` returns a promise that rejects with `"Cancelled"` when `cancel()` is called on a pending timer. Neither call site awaits the result, so without a `.catch(() => {})` every cleanup logs `Uncaught (in promise) Cancelled` to the console. Both call sites swallow the rejection explicitly and document why in a comment.

`bun scripts/static-checks.ts src/lib/components/customer-support/threads-overview.svelte src/lib/components/customer-support/feedback-widget.svelte` passes.

PR #414, opened 2026-05-27
fix(support): stop widget render loop from useDebounce in $effect

stickerdaniel:
The customer support widget froze the whole page with `effect_update_depth_exceeded` as soon as the overview listed a conversation, and on the `?support=open&thread=` deep link the chat never opened. Separately, `bun run dev` 500'd on every SSR page in this repo after the Vite 8 upgrade.

The avatar-preload fallback in `threads-overview.svelte` and the rate-limit auto-clear in `feedback-widget.svelte` (both added in #402) called a runed `useDebounce` function from inside an `$effect`. The effect tracked the debouncer's internal reactive state, so it re-ran, cancelled itself in cleanup, and rescheduled in a loop until Svelte bailed out. The debounced callback never fired, so `allImagesLoaded` never settled. Both now use a plain `setTimeout` with `clearTimeout` cleanup, which keeps the one-shot fallback behaviour without the reactive feedback loop.

The dev SSR 500 (from #288's Vite 7 to 8 bump) came from `isDev()` reading `import.meta.env` off an intermediate variable; Vite 8's dev module runner rejects dynamic `import.meta.env` access. It now reads `import.meta.env.DEV` directly so Vite can statically replace it, with an inline cast so the module still type-checks from the Convex tsconfig.

Reproduced locally by seeding a support thread and opening the widget: the looping effect ran 1002 times before the fix and 0 after. The overview and the `?support=open&thread=` deep link both render and open the conversation with a clean console. `bun scripts/static-checks.ts` and `bun run check:convex` pass.

### DIFF
--- a/src/lib/components/customer-support/feedback-widget.svelte
+++ b/src/lib/components/customer-support/feedback-widget.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
 	import { useConvexClient, useQuery } from '@mmailaender/convex-svelte';
 	import { page } from '$app/state';
-	import { useDebounce, watch } from 'runed';
+	import { watch } from 'runed';
 	import { api } from '$lib/convex/_generated/api';
 	import { supportThreadContext } from './support-thread-context.svelte';
 	import { lockscroll } from '@svelte-put/lockscroll';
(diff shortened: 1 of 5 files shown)


stickerdaniel:
* **#414** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/414?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/414?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #415, opened 2026-05-27
feat(marquee): replace logo cloud with spell marquee

stickerdaniel:
The hero logo cloud used a one-off marquee in src/blocks/magic that only scrolled left, hardcoded its speed, and left fade edges entirely to the parent. It was not reusable anywhere else and had no shared API for direction, duration, or pause-on-hover.

This swaps it for a reusable spell marquee under src/lib/components/spell/marquee. The component supports horizontal and vertical direction, configurable duration, optional mask-based fade edges with an adjustable amount, and pause-on-hover. The registry source ships without a reduced-motion guard, so I added a prefers-reduced-motion fallback that stops the loop for users who opt out of motion.

The hero is migrated to the new component. Its existing gradient and progressive-blur edges are kept (the components own fade is disabled to avoid double-fading), the 16s scroll speed is preserved, and the logos are wrapped in a flex row with a 3rem gap plus trailing padding so spacing stays even across the seam between the duplicated segments.

bun scripts/static-checks.ts on the changed files passes (format, lint, types all green).


stickerdaniel:
* **#415** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/415?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/415?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #416, opened 2026-05-27
chore(lint): ban useDebounce() inside Svelte runes

stickerdaniel:
The support widget froze with `effect_update_depth_exceeded` (fixed in #414) because a runed `useDebounce(...)` result was called inside an `$effect`. `useDebounce` keeps its timer state in an internal `$state` that the returned function reads and writes on every call, so the rune captures it as a dependency and then invalidates it, looping until Svelte bails out. This adds an automated guard so that exact footgun can never reach `main` again.

The new ESLint rule `local/no-debounce-in-rune` errors when a `useDebounce(...)`-bound identifier is called (including `.cancel()`) inside `$effect`, `$effect.pre`, `$effect.root`, `$derived`, or `$derived.by`. It does not flag the safe patterns: the `Debounced` class (reactive debounced value), a `useDebounce` result called from an event handler, or a plain `setTimeout` inside an effect. It applies to `src/**/*.svelte` and `src/**/*.svelte.ts`.

The guard lives in ESLint rather than oxlint on purpose: oxlint JS plugins are in alpha and list "Custom file formats and parsers (e.g. Svelte, Vue, Angular)" under "Not supported yet", so a custom oxlint rule cannot see `.svelte` files where this pattern occurs. AGENTS.md notes a TODO to port it to a local oxlint plugin once oxlint adds Svelte support.

`bun run test:unit` passes (395 tests, including 7 new cases for this rule). `eslint` errors on a `.svelte` fixture that reintroduces the bug and stays clean on the fixed widgets and the existing `new Debounced(...)` usages. `bun scripts/static-checks.ts` passes.

See also: #414

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -548,10 +548,11 @@ How to add: create rule in `eslint/rules/`, register in `eslint.config.js`, add 
 
 #### "This pattern must never appear in code"
 
-Examples: hardcoded aria-labels, barrel icon imports, deprecated Tailwind tokens, bare `animate-spin`.
+Examples: hardcoded aria-labels, barrel icon imports, deprecated Tailwind tokens, bare `animate-spin`, calling a runed `useDebounce()` result inside a `$effect`/`$derived` rune (loops with `effect_update_depth_exceeded`).
 
 → **ESLint custom rule** (for AST-level patterns) or **banned pattern** in `scripts/static-checks.ts` (for simple string matching).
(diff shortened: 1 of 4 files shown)


stickerdaniel:
* **#416** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/416?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/416?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #418, opened 2026-05-28
feat(marquee): add repeat prop for short content

stickerdaniel:
The spell marquee renders the children twice and loops via translateX(-50%). That stays seamless only when a single segment is at least as wide as its container. With short content (a handful of tags, a tiny headline) in a wide container the second segment scrolls off before the next cycle starts, leaving empty space that snaps back at the loop reset. This is the limitation flagged in the original review as finding 5 and deferred.

Add a repeat prop (default 1, identical to current behavior) that renders the children N times inside each segment. A consumer with short content sets it high enough that one segment exceeds the container, restoring seamlessness without any measurement or JS. The math: segment width becomes N times one copy, scroller width 2N copies, translate stays -50% so the loop point still lands seg2 exactly where seg1 started.

bun scripts/static-checks.ts on the change passes (format, lint, types). Seamlessness will be verified on the preview deployment via Playwright coverage sampling across a full cycle once CI is green.

### DIFF
--- a/src/lib/components/spell/marquee/marquee.svelte
+++ b/src/lib/components/spell/marquee/marquee.svelte
@@ -13,6 +13,32 @@
 		direction?: MarqueeDirection;
 		fade?: boolean;
 		fadeAmount?: number;
+		/**
+		 * How many copies of the children to render inside each segment of
+		 * the loop. The marquee stays seamless only when one segment fully
+		 * spans its container along the scroll axis; with short content in
+		 * a long container the loop exposes empty space at the seam.
+		 *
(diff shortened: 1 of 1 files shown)


stickerdaniel:
* **#418** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/418?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/418?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #419, opened 2026-05-28
fix(marketing): keep markdown responses out of shared caches

stickerdaniel:
The CF Workers preview E2E suite has been intermittently failing on the public-a11y and public-agent-surface specs with errors that look like rendering bugs (no `<title>`, no `<html lang>`, `text/markdown` returned for browser navigation). The real cause is at the edge: the marketing markdown response is served with `public, max-age=0, s-maxage=300`, and CF Edge ignores `Vary: Accept`, so the markdown response gets URL-keyed at the edge and is then returned for subsequent HTML requests on the same URL until the TTL expires. This was reproduced live against the current preview: a request with `Accept: text/markdown` followed by a request with `Accept: text/html` on the same URL returns markdown with `cf-cache-status: HIT`. The same vector applies in production behind a custom domain, since shared caches generally ignore Vary.

Switch markdown responses to `private, max-age=300, stale-while-revalidate=300` so they stay out of shared caches entirely; browsers may still cache them per session. The static-URL helpers for `llms.txt`, `robots.txt`, and `sitemap.xml` keep `MARKETING_CACHE_CONTROL` since each of those has a single content type per URL and no cross-type cache key. Drop the AGENTS.md bullet about a preview-only `s-maxage` skip that was never implemented and replace it with a short note about markdown staying out of shared caches. Add regression assertions that the markdown response is `private` and never contains `s-maxage`.

`bun scripts/static-checks.ts` passes. `bun run test:unit -- src/lib/markdown/marketing.test.ts` passes (10/10).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -510,8 +510,9 @@ Marketing routes that are NOT prerendered get edge caching via `handleCacheContr
 
 - Condition: unauthenticated + matches `matchPublicMarketingRoute()`
 - Headers: `Cache-Control: public, s-maxage=3600, stale-while-revalidate=86400`
-- Placed AFTER `handleMarketingMarkdown` in `sequence()` to preserve markdown's own 5-minute TTL
-- **Disabled on CF Workers preview deployments** — preview hostnames (`ALIAS-saas-starter.*.workers.dev`) skip `s-maxage` to prevent stale cached HTML from breaking `Accept: text/markdown` content negotiation.
+- Placed AFTER `handleMarketingMarkdown` in `sequence()` to preserve markdown's own TTL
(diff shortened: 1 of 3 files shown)


stickerdaniel:
* **#419** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/419?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/419?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #420, opened 2026-05-28
fix(marquee): coerce numeric props and pause on focus

stickerdaniel:
Three polish items carried back from the review feedback on the upstream port of this component at SikandarJODD/cnblocks#97.

`safeDuration`, `clampedFadeAmount` and `safeRepeat` now coerce their input through `Number()` before the finite-number check, so a numeric-string prop value (e.g. `repeat="4"`) behaves the same as a numeric one rather than silently falling back to the default.

The two segment `{#each}` blocks iterate `{ length: safeRepeat }` instead of `Array(safeRepeat)`. The array-like literal is cheaper than allocating a sparse array on every reactivity tick and matches the idiom we used in the earlier magic marquee.

The `pauseOnHover` rule now also fires on `.group:focus-within`, so when a marquee carries focusable children the animation pauses while the user is interacting via keyboard, not just hovering.

No API changes, defaults unchanged. bun scripts/static-checks.ts on the file passes.

### DIFF
--- a/src/lib/components/spell/marquee/marquee.svelte
+++ b/src/lib/components/spell/marquee/marquee.svelte
@@ -59,14 +59,19 @@
 	}: MarqueeProps = $props();
 
 	const isVertical = $derived(direction === 'up' || direction === 'down');
-	const safeDuration = $derived(Number.isFinite(duration) ? Math.max(duration, 0.01) : 20);
-	const clampedFadeAmount = $derived(
-		Number.isFinite(fadeAmount) ? Math.min(Math.max(fadeAmount, 0), 50) : 10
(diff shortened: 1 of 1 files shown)


stickerdaniel:
* **#420** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/420?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/420?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #424, opened 2026-05-29
fix(settings): render active tab from page.url for flash-free SSR

### LINKED ISSUE #413: Settings tab selection (useSearchParams) flashes default tab on SSR
### What happens

Deep-linking or reloading the account settings page on a non-default tab flashes the default `account` tab before switching. Visit `/en/app/settings?tab=email` (or `?tab=password`) and the `account` tab renders first, then jumps to the requested tab after hydration.

### Why

`src/routes/[[lang]]/app/settings/+page.svelte` drives the active tab with `useSearchParams` (`runed/kit`). That helper initialises its cache from the URL **client-only**: `node_modules/runed/dist/utilities/use-search-params/use-search-params.svelte.js` `#initializeCacheFromURL()` returns early when `!BROWSER`. So on the server `searchParams.tab` resolves to the schema default (`account`), the SSR HTML renders the account tab, and the client corrects to the real tab after hydration. Anything where `useSearchParams` drives an SSR-visible selection has the same flash.

### Fix

Read the param from `page.url.searchParams` (SvelteKit populates `page.url` on the server, so SSR is correct and there is no flash) and write with `goto(resolve(...), { replaceState: true })`. This is the pattern `src/routes/[[lang]]/admin/support/+page.svelte` already uses for thread selection, which does not flash:

```ts
import { page } from '$app/state';
import { goto } from '$app/navigation';
import { resolve } from '$app/paths';

const activeTab = $derived(parseTab(page.url.searchParams.get('tab')));

function setTab(value: string) {
  const url = new URL(page.url);
  if (value === DEFAULT_TAB) url.searchParams.delete('tab');
  else url.searchParams.set('tab', value);
  goto(resolve(url.pathname + url.search), { replaceState: true, keepFocus: true, noScroll: true });
}
```

`useSearchParams` is fine where the selection is not server-rendered (e.g. client-loaded tables behind a skeleton), but not for SSR-visible view selection.

stickerdaniel:
Closes #413

The settings page used `useSearchParams` from `runed/kit` to drive tab selection, which initialises its cache from the URL only when `BROWSER` is true. Deep-linking `/en/app/settings?tab=email` rendered the `account` tab on the server and jumped after hydration. Any SSR-visible view selection backed by `useSearchParams` had the same flash.

The settings page now reads the active tab from `page.url.searchParams.get('tab')` in a `\$derived`, the same pattern `admin/support` thread selection already uses. SvelteKit populates `page.url` during SSR, so the right tab renders on first paint with no client-side correction. Writes go through `goto(...)` with `keepFocus` and `noScroll`, and the default `account` is stripped from the URL so links stay clean. \`AGENTS.md\` now documents the URL vs cookie vs localStorage decision for client-set view state under Data-fetching patterns, and adds a Prerendering-constraints note that request-time values (e.g. the \`sidebar_state\` cookie) must be read in \`hooks.server.ts\` into \`event.locals\` rather than \`cookies.get\` in a shared layout load.

\`bun scripts/static-checks.ts 'src/routes/[[lang]]/app/settings/+page.svelte' AGENTS.md\` passes. Manual: hard-reload \`/en/app/settings?tab=email\` shows the email tab on first paint with no flash.


stickerdaniel:
* **#424** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/424?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/424?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #426, opened 2026-05-29
fix(prompt-input): re-measure textarea on resize

### LINKED ISSUE #425: PromptInputTextarea latches a tall height and never recovers
## Problem

`PromptInputTextarea` auto-grows via a JS `scrollHeight` watch that writes an explicit pixel `style.height`. Two things make it latch a wrong height:

1. It only resets `height = 'auto'` before measuring **when `scrollTop === 0`**. While the field is scrolled, the reset is skipped, so `scrollHeight` is measured against the already-tall `clientHeight` and the box can only ever grow.
2. `scrollHeight` is **width-dependent**. The base `ui/textarea` has `field-sizing: content` + a wrapping placeholder, so when the composer sits in a resizable/narrow pane the empty placeholder wraps onto many lines and `scrollHeight` is large. The watch only re-runs on value/maxHeight changes, never on width changes, so once a tall height is written at a narrow width it sticks after the pane widens again.

Net effect: an **empty** composer can end up stuck at `maxHeight` (e.g. 240px) at a normal width. Reproduced in a fork with a resizable chat/artifact split: measure while narrow -> 240px -> widen -> still 240px.

Note: dropping the JS in favour of `field-sizing: content` alone is not enough, because Firefox still does not support `field-sizing` (not Baseline as of 2026), so the JS path must stay for cross-browser auto-grow.

## Fix

- Always reset `height = 'auto'` before measuring (drop the `scrollTop === 0` guard) so the field can shrink as well as grow.
- Add a `ResizeObserver` on the textarea so the height re-measures on width changes and tracks the current width instead of latching.

Keeps JS-driven auto-grow on all browsers (incl. Firefox); no interaction change on Chrome/Safari beyond the bug disappearing. PR incoming.

stickerdaniel:
Closes #425

The composer's `PromptInputTextarea` auto-resize could latch a tall height it never recovered from: an empty composer would sit stuck at `maxHeight` (240px) even at a normal width. It happened because the JS auto-resize writes an explicit pixel height from a width-dependent `scrollHeight`, only reset `height` to `auto` when `scrollTop === 0`, and only re-ran on value or maxHeight changes. So a height measured while the field was narrow (the base textarea uses `field-sizing: content` and its placeholder wraps onto many lines in a narrow or resizable pane) was written once and stuck after the pane widened again.

This always resets `height` to `auto` before measuring so the field can shrink as well as grow, and adds a `ResizeObserver` so the height re-measures whenever the width changes and tracks the current width instead of latching. The JS path stays (rather than relying on `field-sizing` alone) because Firefox still does not support `field-sizing`, so dropping it would remove cross-browser auto-grow.

Verified in a downstream fork against a live resizable chat/artifact split: an empty composer forced narrow grows to 240px as before, and on widening recovers to 44px instead of sticking. Typing-to-grow, the maxHeight cap with overflow, and clearing all behave as before; no interaction change on Chrome/Safari beyond the bug disappearing.


stickerdaniel:
* **#426** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/426?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/426?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #429, opened 2026-05-29
feat(dev): add OpenRouter model capability eval

stickerdaniel:
Adds a local tool, `bun run model:eval`, for checking whether an OpenRouter chat model supports everything the app asks of its chat models before swapping `CHAT_MODEL_ID`. It covers the real requirements: streaming text, reasoning traces, image input, PDF input, and tool calling. Each capability is streamed through the app's own message-materialization pipeline (the same `deriveUIMessagesFromTextStreamParts` to `transformToDisplayMessage` path the chat UI uses), so a pass means the output actually renders as the right reasoning, tool, and text parts, not just that bytes arrived. It prints a pass/fail capability matrix, defaults to the model in use, and takes `--model` (repeatable) to check candidates.

Not CI and not wired into any build. It burns OpenRouter credits and needs `OPENROUTER_API_KEY` (env or `.env.convex.local`). The image and PDF fixtures are committed binaries, so there is no runtime image tooling.

Verified against `google/gemini-2.5-flash-lite` (all six capabilities pass) and the current `CHAT_MODEL_ID` `google/gemma-4-26b-a4b-it`, which surfaces a real gap: it advertises no file modality and hallucinates a doc-reader tool call instead of reading the PDF, so the pdf check fails even though the app accepts PDF uploads. `bun scripts/static-checks.ts` passes.


stickerdaniel:
* **#429** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/429?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/429?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #282, opened 2026-04-05
style(ui): Add active:translate-y-px press effect to all text links

stickerdaniel:
Auth pages, error display, breadcrumb, admin support, and team
block links were missing the press-down effect that buttons and
sidebar items already have.

PR #431, opened 2026-05-29
fix(ui): correct text-link press effect

### LINKED ISSUE #411: Breadcrumb press effect (active:translate-y-px) does nothing on first crumb
## Summary

The `active:translate-y-px` press effect on breadcrumb links does nothing on the **first** breadcrumb crumb (the root item), while it works on every following crumb. The first crumb stays visually static when pressed.

## Root cause

`authenticated-header.svelte` renders the first breadcrumb item with a responsive-hide class:

```svelte
<Breadcrumb.Item class={index === 0 ? 'hidden md:block' : ''}>
```

`Breadcrumb.Item` defaults to `inline-flex items-center gap-1.5`. The `hidden md:block` override replaces `inline-flex` with `block` (tailwind-merge keeps the later display utility), so at `md+` the `<li>` is `display: block` instead of a flex container.

Because the `<li>` is no longer a flex container, its child `<a data-slot="breadcrumb-link">` keeps its default `display: inline`. CSS `translate` (and `transform`) is ignored on non-replaced inline elements, so `active:translate-y-px` has no effect on the first crumb.

Every other breadcrumb item keeps the default `inline-flex`, which blockifies its child link to `display: block`, so the press effect works there.

## Secondary issue

`breadcrumb-link.svelte` uses `transition-colors`:

```svelte
class: cn('hover:text-foreground transition-colors active:translate-y-px', className),
```

`transition-colors` does not include `translate` in its `transition-property`, so even on the crumbs where the press works, the 1px shift snaps with no easing instead of transitioning like other press targets in the app (e.g. sidebar buttons that use `transition-transform`, which in Tailwind v4 includes `translate`).

## Reproduction

1. Open any authenticated page with at least two breadcrumb crumbs at `md+` width.
2. Press and hold the first/root crumb. It does not move.
3. Press and hold a later crumb. It shifts down 1px.

## Suggested fix

Keep the first item a flex container so the inner link is blockified:

```svelte
<Breadcrumb.Item class={index === 0 ? 'hidden md:inline-flex' : ''}>
```

Optionally, transition the press smoothly in `breadcrumb-link.svelte`:

```svelte
class: cn('hover:text-foreground transition-[color,translate] active:translate-y-px', className),
```

## Affected files

- `src/lib/components/authenticated/authenticated-header.svelte` (first breadcrumb item class)
- `src/lib/components/ui/breadcrumb/breadcrumb-link.svelte` (transition property)

### LINKED ISSUE #412: Remove dead active:translate-y-px from inline text links (no-op on display:inline)
## Summary

The `active:translate-y-px` press effect added in #282 ("to all text links") is a no-op on every text link that computes to `display: inline`. CSS `translate`/`transform` is ignored on non-replaced inline elements, so the press does nothing on bare inline `<a>` links inside sentences. This is the same root cause as #411, but #411 is specifically the breadcrumb first crumb (fixed by making the item a flex container). This issue tracks the inline text links, where the right fix is to drop the dead class rather than force a display change.

A 1px downward shift on a single word inside a sentence shifts it against the surrounding baseline, which reads as jitter rather than a press, so the effect is undesirable on these links anyway. `inline-block` is not a good fix because it stops the link wrapping across lines.

## Scope

Remove `active:translate-y-px` from the inline text links below. Leave it on links that already sit in a `flex`/`block`/`inline-block` context (those work and are intentional).

- [ ] `src/lib/components/ErrorDisplay.svelte` (the "back home" link in the `Empty.Description`)
- [ ] `src/routes/[[lang]]/(auth)/signin/+page.svelte` (terms, privacy, back-home footer links)
- [ ] `src/routes/[[lang]]/(auth)/signup/+page.svelte` (sign-in cross-link, plus terms, privacy, back-home footer links)
- [ ] `src/routes/[[lang]]/(auth)/signin/SignInForm.svelte` (the sign-up cross-link only; keep the forgot-password link, it is in a `flex` row and works)
- [ ] `src/routes/[[lang]]/(auth)/signin/SignUpForm.svelte` (sign-in cross-link)
- [ ] `src/routes/[[lang]]/(auth)/reset-password/+page.svelte` (sign-in link in the success message, back-to-signin link, and terms/privacy/back-home footer links)
- [ ] `src/routes/[[lang]]/(auth)/forgot-password/+page.svelte` (back-to-signin link, plus terms/privacy/back-home footer links)

## Not in scope (verified working, leave as is)

`active:translate-y-px` works on these because the element is not `display: inline`:

- `src/blocks/team/team-two.svelte` (`inline-block`)
- `sidebar-menu-button.svelte`, `sidebar-menu-sub-button.svelte`, `sidebar-thread-list.svelte`, `authenticated-sidebar.svelte` (button / flex)
- `ChatAttachments.svelte`, `ReasoningTrigger.svelte`, `admin/support/thread-details.svelte` (flex)
- `SignInForm.svelte` forgot-password link (flex row)

The breadcrumb component is tracked separately in #411.

## Suggested guard

To stop this regressing, consider only applying the press utility to links that are block/flex/inline-block, or add a lint note that `active:translate-y-px` on a bare inline `<a>` is a no-op.

stickerdaniel:
The `active:translate-y-px` press effect from #282 is ignored by CSS on `display: inline` elements, so it silently did nothing on bare inline text links, and on the breadcrumb it worked on every crumb except the first.

For breadcrumbs, the first crumb was rendered with `hidden md:block`, which made the `<li>` `display: block` and left its child link `display: inline`, so the press never fired there while it worked on the later (inline-flex) crumbs. The item now uses `md:inline-flex` so the link is blockified, and `breadcrumb-link.svelte` transitions `translate` alongside color so the 1px shift eases instead of snapping. For inline sentence links on the auth pages and the error page back-home link, the effect is a no-op on `display: inline` and reads as baseline jitter where it does apply, so the class is dropped. It is kept on the flex/block/inline-block targets (sidebar buttons, attachments, reasoning trigger, the forgot-password flex link) where it works as intended.

Closes #411
Closes #412

`bun scripts/static-checks.ts` passes on all changed files (lint, format, svelte-check 0 errors).

### DIFF
--- a/src/lib/components/ErrorDisplay.svelte
+++ b/src/lib/components/ErrorDisplay.svelte
@@ -92,10 +92,7 @@
 			</InputGroup.Root>
 			<Empty.Description>
 				{translations.error_page.need_help}
-				<a
-					href={resolve(homeHref)}
-					class="text-primary underline-offset-4 hover:underline active:translate-y-px"
-				>
+				<a href={resolve(homeHref)} class="text-primary underline-offset-4 hover:underline">
 					{translations.error_page.back_home}
 				</a>
 			</Empty.Description>
--- a/src/lib/components/authenticated/authenticated-header.svelte
+++ b/src/lib/components/authenticated/authenticated-header.svelte
@@ -36,7 +36,7 @@
 					{#if index > 0}
 						<Breadcrumb.Separator class="hidden md:block" />
 					{/if}
-					<Breadcrumb.Item class={index === 0 ? 'hidden md:block' : ''}>
+					<Breadcrumb.Item class={index === 0 ? 'hidden md:inline-flex' : ''}>
 						{#if item.isLast}
 							<Breadcrumb.Page>{item.label}</Breadcrumb.Page>
 						{:else}
--- a/src/lib/components/ui/breadcrumb/breadcrumb-link.svelte
+++ b/src/lib/components/ui/breadcrumb/breadcrumb-link.svelte
@@ -16,7 +16,10 @@
 
 	const attrs = $derived({
 		'data-slot': 'breadcrumb-link',
-		class: cn('hover:text-foreground transition-colors active:translate-y-px', className),
+		class: cn(
+			'hover:text-foreground transition-[color,translate] active:translate-y-px',
+			className
+		),
 		href,
 		...restProps
 	});
--- a/src/routes/[[lang]]/(auth)/forgot-password/+page.svelte
+++ b/src/routes/[[lang]]/(auth)/forgot-password/+page.svelte
@@ -222,7 +222,7 @@
 								<a
 									href={resolve(localizedHref('/signin'))}
 									data-testid="forgot-password-back-link"
-									class="underline underline-offset-4 active:translate-y-px"
+									class="underline underline-offset-4"
 								>
 									<T keyName="auth.forgot_password.back_to_signin" defaultValue="Back to sign in" />
 								</a>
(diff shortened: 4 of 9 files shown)


stickerdaniel:
* **#431** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/431?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/431?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #432, opened 2026-05-29
fix(vite): exclude blocked ports from maxAttempts

stickerdaniel:
The dev port allocator skips WHATWG bad-fetch ports so the chosen port stays reachable, but a skipped port still consumed one of the maxAttempts iterations. That made the limit reflect ports walked rather than ports actually probed, so a small maxAttempts landing on a cluster of blocked ports could give up before checking any connectable port.

This counts only non-blocked ports toward the limit: blocked ports are skipped for free via a separate checked counter, while the scan offset keeps advancing. The loop still terminates because the blocked set is finite (max 10080), so once the offset passes it every port counts. Same change applied upstream in juliusmarminge/agent-tools#28, where this allocator originates.

bun scripts/static-checks.ts vite.config.ts passes.

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -42,9 +42,13 @@ const BAD_FETCH_PORTS = new Set([
 ]);
 
 async function findAvailablePort(startPort: number, maxAttempts = 100): Promise<number> {
-	for (let attempt = 0; attempt < maxAttempts; attempt += 1) {
-		const port = startPort + attempt;
+	// Blocked ports don't count toward maxAttempts, so the limit always reflects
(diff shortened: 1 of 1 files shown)


stickerdaniel:
* **#432** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/432?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/432?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #433, opened 2026-05-29
ci(fork-preview): gated E2E + preview deploy for fork PRs

stickerdaniel:
Fork PRs from external contributors cannot currently be merged without an admin override. Branch protection on `main` requires the `Workers Builds: saas-starter` and `E2E Tests` checks, but Cloudflare Workers Builds does not build fork PRs, so neither check ever runs for a fork and the merge button stays blocked.

This adds a `pull_request_target` workflow that runs only for fork PRs and is gated behind the `fork-preview` GitHub Environment (required reviewer). Nothing deploys or touches the preview-scoped secrets until a maintainer approves the run, so external contributors cannot spam preview builds. After approval the job checks out the reviewed fork SHA, builds a Cloudflare Workers preview in CI by reusing the existing `scripts/deploy.ts` and `scripts/cf-deploy.ts` (so the fork path stays in sync with the production CF path), runs Playwright against the live preview, and posts both required commit statuses. Same-repo PRs are unaffected and keep using CF Workers Builds plus `e2e-preview-cf.yml`.

Security choices: `pull_request_target` is required because fork `pull_request` jobs get no secrets and a read-only token; the workflow file therefore comes from `main` and cannot be modified by a fork. The checkout uses `persist-credentials: false`, deploy and test secrets are scoped to their steps (so `bun install` postinstall runs without secrets), the preview key is `fork-pr-<number>` to avoid cross-fork alias collisions, and only preview-scoped Convex and Cloudflare credentials are exposed. Approving a run means trusting the entire checked-out fork tree to run with those preview secrets, which is the documented trust boundary.

This required two repo-config changes already applied: the `fork-pr-contributor-approval` policy is set to `all_external_contributors`, and the `Workers Builds: saas-starter` required check was un-pinned from the Cloudflare app so the Actions-posted status can satisfy it.

The workflow cannot be exercised by this PR (for forks GitHub uses the workflow file from the base branch), so it is validated after merge by driving a fork PR through the approval gate. `bun scripts/static-checks.ts` passes on the workflow file.

### DIFF
--- a/.github/workflows/e2e-fork-preview.yml
+++ b/.github/workflows/e2e-fork-preview.yml
@@ -1,0 +1,150 @@
+name: E2E Tests (Fork Preview)
+
+# Builds a Cloudflare Workers preview and runs Playwright E2E for FORK pull
+# requests, gated behind the `fork-preview` Environment (required reviewer).
+# Cloudflare Workers Builds does not build fork PRs, and a fork `pull_request`
+# gets no secrets / a read-only token, so this uses `pull_request_target`
+# (base-repo context: secrets + read-write token) and only runs after a
+# maintainer approves the `fork-preview` deployment. Approving means trusting
+# the entire checked-out fork tree (lockfile, install/postinstall scripts,
+# scripts/deploy*, tests) to run with the preview-scoped secrets.
+#
+# Same-repo PRs are skipped here (they use CF Workers Builds + e2e-preview-cf.yml).
+#
+# NOTE: Convex preview-quota recovery is intentionally disabled (no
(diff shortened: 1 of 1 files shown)


stickerdaniel:
* **#433** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/433?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/433?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #434, opened 2026-05-29
fix(admin): inline passkey icon to survive ad blockers

### LINKED ISSUE #409: admin/users route crashes in dev when a privacy/ad blocker is active (lucide fingerprint icon path blocked)
## Problem

The admin users table route (`/[[lang]]/admin/users`) fails to hydrate and the page crashes for any user whose browser has a privacy or ad blocker enabled (uBlock Origin, Brave Shields, Privacy Badger, etc.).

Console shows:

```
GET .../node_modules/@lucide/svelte/dist/icons/fingerprint.js net::ERR_BLOCKED_BY_CLIENT
Failed to fetch dynamically imported module: .../nodes/24.js
```

## Root cause

`src/routes/[[lang]]/admin/users/provider-badge.svelte` imports the passkey badge icon via:

```ts
import Fingerprint from '@lucide/svelte/icons/fingerprint';
```

In dev, Vite serves the unbundled module graph, so this resolves to a real fetchable URL containing the literal word `fingerprint`. Anti-tracking filter lists block any request whose path contains `fingerprint`, returning `ERR_BLOCKED_BY_CLIENT`. The blocked module is a transitive dependency of the route chunk, so the dynamic import of the route node fails and SvelteKit cannot hydrate the page.

## Scope

This is effectively dev-only. In a production build the static import is bundled into a content-hash-named chunk, so no URL path contains `fingerprint` and the blocker has nothing to match. Verified against a production client build: no output filename contains `fingerprint`. But it fully breaks local development for anyone running a blocker.

## Suggested fix

Inline the fingerprint glyph as an SVG, the same way the Google and GitHub provider icons are already inlined in that file. This removes the network request entirely, so there is no path for a blocker to match, and it is identical in dev and prod. The Lucide `mail` icon for the credential provider is unaffected and can stay as an import.

stickerdaniel:
Closes #409

The `/admin/users` route failed to hydrate and the page crashed for anyone browsing with a privacy or ad blocker enabled. The passkey provider badge imported its icon via `import Fingerprint from '@lucide/svelte/icons/fingerprint'`. In dev, Vite serves lucide from the unbundled module graph, so the import resolves to a real fetchable URL containing the literal word `fingerprint`. Anti-tracking filter lists block any request whose path matches `fingerprint`, returning `ERR_BLOCKED_BY_CLIENT`, and since that module is a transitive dependency of the route chunk, the route node import fails and SvelteKit cannot hydrate the page.

Renaming the import binding does not help, because the blocked thing is the module URL path, not the JS identifier, and a local re-export still fetches the same module by URL in dev. Instead I inlined the fingerprint glyph as an SVG, the same way the Google and GitHub provider icons are already inlined in this file. That removes the network request entirely, so there is nothing for a blocker to match, and the rendering is identical in dev and prod. The path data and stroke attributes are taken verbatim from the lucide source; fingerprint is a multi-path stroke icon, so the SVG uses `fill="none"` with `stroke="currentColor"` rather than the filled-path style of the Google/GitHub badges. The `mail` lucide import for the credential provider is unaffected and stays as an import.

`bun scripts/static-checks.ts` passes on the changed file (format, lint, types, banned patterns).

### DIFF
--- a/src/routes/[[lang]]/admin/users/provider-badge.svelte
+++ b/src/routes/[[lang]]/admin/users/provider-badge.svelte
@@ -1,7 +1,6 @@
 <script lang="ts">
 	import { T, getTranslate } from '@tolgee/svelte';
 	import Mail from '@lucide/svelte/icons/mail';
-	import Fingerprint from '@lucide/svelte/icons/fingerprint';
 
 	const { t } = getTranslate();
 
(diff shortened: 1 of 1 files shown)


stickerdaniel:
* **#434** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/434?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/434?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #435, opened 2026-05-29
fix(sidebar): persist collapse state across reloads

### LINKED ISSUE #404: Authenticated sidebar: open state stored in a write-only cookie, should be PersistedState
Sidebar open/collapsed state does not survive a reload across all authenticated routes (`/app/*` and `/admin/*` via `AuthenticatedLayout`).

`Sidebar.Provider` at `src/lib/components/ui/sidebar/sidebar-provider.svelte:14` defaults `open = $bindable(true)` and `AuthenticatedLayout` never passes `open=`, so it always starts open. The provider writes `sidebar:state` to `document.cookie` on every toggle (`sidebar-provider.svelte:34`), but nothing reads it back, not the server (no `cookies.get('sidebar:state')` in `+layout.server.ts` or `hooks.server.ts`) and not the client. The write is dead.

Adjacent in the same tree, `authenticated-sidebar.svelte:26` already uses `new PersistedState('ai-chat-threads-open', true)` for the chat threads accordion. Same pattern would work for the main sidebar open state and would drop the cookie path entirely.

stickerdaniel:
Closes #404

The authenticated sidebar's open/collapsed state did not survive a reload on `/app/*` or `/admin/*`. `Sidebar.Provider` wrote the state cookie to `document.cookie` on every toggle, but nothing ever read it back, so the shell always rendered open and reflowed full-width on reload for anyone who kept it collapsed.

This wires up the missing read side with shadcn's canonical SvelteKit pattern. A server hook reads the cookie into `event.locals.sidebarOpen`, the root layout load returns it, and `AuthenticatedLayout` forwards it to `Sidebar.Provider` so SSR paints the correct state on first load. Reading in the hook rather than a `+layout.server.ts` load keeps the prerendered marketing pages buildable, the same channel the JWT token already flows through. The cookie is also renamed from `sidebar:state` to `sidebar_state` so the name is a valid cookie token and the write and read sides share the single `SIDEBAR_COOKIE_NAME` constant.

A Playwright spec collapses the sidebar, reloads, and asserts it comes back collapsed, then re-expands and reloads to cover the reverse, guarding the read path against regressing to a write-only cookie. The view-state decision tree in `AGENTS.md` also gets a one-line quick rule for choosing between URL, cookie + SSR, and `PersistedState`.

`bun run build` prerenders cleanly, `svelte-check` passes with 0 errors, and the new e2e spec passes locally.


stickerdaniel:
* **#435** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/435?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/435?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #438, opened 2026-05-30
feat(sidebar): support action callback on nav items

### LINKED ISSUE #408: Sidebar NavItem should support an action callback, not just URL navigation
The authenticated sidebar (`src/lib/components/authenticated/authenticated-sidebar.svelte`) only renders nav items as anchors that navigate to `item.url`. The `NavItem` type (`src/lib/components/authenticated/types.ts`) requires `url: string`, so every nav item has to be a link.

Some sidebar entries are actions, not destinations. A "New X" item that opens a creation dialog, a command-palette trigger, or any handler that should run without a route change has no clean home in the sidebar header today. The workaround is a phantom route whose only job is to host a modal, which then needs its own page, redirect, and back-link handling.

Proposal: let a nav item carry a callback instead of a URL.

- Add `onSelect?: () => void` to `NavItem` and make `url?: string` optional.
- In the standard nav-item branch, render `<button type="button" onclick={item.onSelect}>` when `onSelect` is set, otherwise keep the existing `<a href={resolve(item.url)}>`. Extract the shared item body (icon, label, kbd) into a snippet so the button and anchor do not duplicate markup.
- Guard the `resolve(item.url)` reads for the now-optional url.

Backward compatible: configs that set `url` are unaffected, and `<button>` is the correct semantics for a control that triggers an action rather than navigating. Happy to send a PR.

stickerdaniel:
Closes #408

The authenticated sidebar rendered every nav item as an anchor to `item.url`, and the `NavItem` type required `url`, so a sidebar entry could only be a link. Entries that are actions rather than destinations (open a creation dialog, trigger the command palette) had no clean home and needed a phantom route whose only job was to host a modal.

This makes `url` optional and adds an `onSelect?: () => void` callback to `NavItem`. In the standard nav-item branch the item renders as an anchor (`resolve(item.url)`) when it has a `url`, otherwise as a `<button type="button">` that runs `onSelect`. The callback is folded into the existing `MenuButton` onclick so the haptic feedback still fires, and branching on `item.url` lets TypeScript narrow the type for a clean `resolve()` with no assertion. The shared icon, label and kbd body is extracted into a snippet so the anchor and button branches do not duplicate markup. Configs that set `url` are unaffected and `FooterLink` stays URL-only.

`bun scripts/static-checks.ts` passes for both files. The sidebar E2E suite (`ai-chat.spec.ts`, `sidebar-persistence.spec.ts`) is green.

### DIFF
--- a/src/lib/components/authenticated/authenticated-sidebar.svelte
+++ b/src/lib/components/authenticated/authenticated-sidebar.svelte
@@ -7,7 +7,7 @@
 	import { resolve } from '$app/paths';
 	import type { ComponentProps } from 'svelte';
 	import { T } from '@tolgee/svelte';
-	import type { NavSubItem, SidebarConfig, User } from './types';
+	import type { NavItem, NavSubItem, SidebarConfig, User } from './types';
 	import { haptic } from '$lib/hooks/use-haptic.svelte';
 	import { PersistedState } from 'runed';
 	import ChevronRightIcon from '@lucide/svelte/icons/chevron-right';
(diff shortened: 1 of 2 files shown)


stickerdaniel:
* **#438** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/438?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/438?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #439, opened 2026-05-30
feat(chat): preview markdown and text attachments

### LINKED ISSUE #410: Attachment preview renders markdown/text files as raw source in an iframe
The attachment preview dialog in `src/lib/chat/ui/ChatAttachments.svelte` renders any non-image attachment with `<iframe src={url}>` (currently around the `{:else if !isImage}` branch). For PDFs this gives the browser PDF viewer, which is fine. For text-based files (markdown, txt, csv, json) the iframe just shows the raw file source as plain text, so a `.md` file displays its unrendered markdown (`# Heading`, `[label](url)`, etc.) instead of formatted content.

The template's support chat only accepts images and PDF today (`ALLOWED_FILE_EXT_MIME` in `src/lib/chat/core/types.ts`), so this does not surface in the default template. It surfaces as soon as a consumer allows `text/markdown` (or other text) attachment types through the shared `ChatAttachments` component, where the raw-text preview looks broken.

## Proposal

In the preview dialog, branch on the mime type:

- Keep the `<iframe>` for PDFs and other browser-native types.
- For `text/markdown` (and ideally `text/plain`), fetch the file text and render it instead of embedding the raw bytes. Markdown can reuse the existing `Response` component (`src/lib/components/ai-elements/response/Response.svelte`) that already renders message markdown with sanitization; plain text can render in a `<pre>`.

## Acceptance

Opening a `.md` attachment in the preview shows formatted markdown (headings, links, lists, code blocks), not the raw source.

stickerdaniel:
Closes #410

The attachment preview dialog rendered every non-image attachment with an iframe, so a markdown or text file showed its raw source (`# Heading`, `[label](url)`) instead of formatted content. It did not surface in the default template because only images and PDF were accepted, but it broke as soon as a consumer allowed text/markdown through the shared `ChatAttachments` component.

Text-based attachments now render through the same Streamdown engine that already renders chat messages: markdown formatted, plain text in a `pre` block, and JSON or code syntax-highlighted, while PDFs keep the browser iframe. The renderer is locked down for untrusted uploaded content so it stays at or below the trust level the app already accepts for message rendering. Raw HTML stays off. Images are blocked through a snippet override rather than the prefix allowlist, because Streamdown still renders relative and protocol-relative URLs (`//evil.com/x.png`) regardless of `allowedImagePrefixes`. Links are clickable only when the resolved href is absolute http(s). Mermaid is shown as a code block instead of a rendered diagram.

Freshly attached files read their text from the in-memory blob. Already-sent attachments, such as an admin viewing a user upload or a reloaded thread, load it through a new `getAttachmentText` action, since Convex storage URLs do not send CORS headers and cannot be read cross-origin from the browser. The action validates the URL against the deployment's own storage origin (`CONVEX_CLOUD_URL`) and the `/api/storage/` path, requires a text content type, and caps the response size.

The template now accepts `.md` and `.txt` uploads in both the client and server allowlists so the preview is reachable, and the paste handler accepts text files whose clipboard MIME type is empty.

In the worktree, `bun run check:convex`, `bun scripts/static-checks.ts` on the changed files, and `bun run test:unit` pass. `bun run test:e2e e2e/ai-chat-attachment-preview.spec.ts` passes: it uploads a `.md`, opens the preview, and asserts the rendered heading with no image loaded.

### DIFF
--- a/e2e/ai-chat-attachment-preview.spec.ts
+++ b/e2e/ai-chat-attachment-preview.spec.ts
@@ -1,0 +1,49 @@
+import { test, expect } from '@playwright/test';
+import { waitForAuthenticated } from './utils/auth';
+
+const MARKDOWN = `# Preview Heading
+
+A paragraph with a [safe link](https://example.com) and an ![tracker](https://example.com/pixel.png).
+
+- first item
+- second item
+`;
+
+test.describe('AI Chat - attachment text preview', () => {
+	test('previewing an attached .md renders formatted markdown, not raw source', async ({
+		page
+	}) => {
+		await page.goto('/app/ai-chat');
+		await waitForAuthenticated(page);
+		await page.waitForURL(/\/app\/ai-chat\?thread=/, { timeout: 15000 });
+		await expect(page.locator('textarea')).toBeVisible({ timeout: 10000 });
+
+		// Attach a markdown file through the hidden file input. Uploading also
+		// exercises the server MIME allowlist: the chip only becomes clickable
+		// once the upload succeeds.
+		await page
+			.locator('input[type="file"]')
+			.first()
+			.setInputFiles({
+				name: 'e2e-notes.md',
+				mimeType: 'text/markdown',
+				buffer: Buffer.from(MARKDOWN, 'utf8')
+			});
+
+		const chip = page.getByTestId('attachment-chip').first();
+		await expect(chip).toBeVisible({ timeout: 15000 });
+		// role="button" is only set once the attachment is clickable (upload done).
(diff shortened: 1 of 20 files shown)


stickerdaniel:
* **#439** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/439?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/439?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #440, opened 2026-05-30
feat(sidebar): reveal shortcuts on hover with fade

stickerdaniel:
The sidebar shortcut hint (a `Kbd.Group` with `ml-auto opacity-0`) reserves horizontal space even while hidden, so a longer nav label competes with the always-reserved shortcut and gets truncated by the menu button's `[&>span:last-child]:truncate`, even when the shortcut is not visible.

This renders the hint absolutely positioned at the right, hidden until hover. The label keeps the full row width when idle and only truncates if it genuinely overflows. On hover the shortcut appears over a short left-to-right gradient that fades the label out underneath it, so the two never overlap. The previously duplicated kbd markup is pulled into a shared `shortcutHint` snippet used by both nav items and footer links, and `relative` is added to the affected menu buttons so the absolute hint positions against the row (which already clips via `overflow-hidden`).

`bun scripts/static-checks.ts src/lib/components/authenticated/authenticated-sidebar.svelte` passes.

### DIFF
--- a/src/lib/components/authenticated/authenticated-sidebar.svelte
+++ b/src/lib/components/authenticated/authenticated-sidebar.svelte
@@ -30,14 +30,26 @@
 	{#if item.icon}
 		<item.icon />
 	{/if}
-	<span><T keyName={item.translationKey} /></span>
+	<span class="min-w-0 truncate"><T keyName={item.translationKey} /></span>
 	{#if item.kbd}
-		<Kbd.Group class="ml-auto opacity-0 group-hover/menu-button:opacity-50">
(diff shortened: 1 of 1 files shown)


stickerdaniel:
* **#440** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/440?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/440?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #441, opened 2026-05-30
feat(sidebar): reveal thread timestamp on hover

stickerdaniel:
Each AI chat thread row always shows a compact relative time ("24d") pinned right with `ml-auto shrink-0`, which permanently steals width from the title and truncates it early (for example "What's the weather like i…").

This hides the timestamp until hover and gives the title the full row width when idle. On hover it reveals a richer, locale-aware absolute date and time (via `Intl.DateTimeFormat`, for example "05/30, 02:35 PM") over a short left-to-right gradient that fades the title out underneath it, so they never overlap. It also drops the hardcoded English `now/m/h/d` units in favor of locale-aware formatting. The timestamp is an absolutely positioned sibling of the menu button inside the already-`relative` sub-item.

`bun scripts/static-checks.ts src/lib/components/authenticated/sidebar-thread-list.svelte` passes.

### DIFF
--- a/src/lib/components/authenticated/sidebar-thread-list.svelte
+++ b/src/lib/components/authenticated/sidebar-thread-list.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
 	import * as Sidebar from '$lib/components/ui/sidebar/index.js';
 	import { resolve } from '$app/paths';
+	import { page } from '$app/state';
 	import { haptic } from '$lib/hooks/use-haptic.svelte';
 	import autoAnimate from '@formkit/auto-animate';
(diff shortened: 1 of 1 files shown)


stickerdaniel:
* **#441** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/441?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/441?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #444, opened 2026-06-01
fix(chat): stop reasoning indicator flicker

stickerdaniel:
The reasoning indicator briefly disappeared between "Connecting" and "Thinking" at the start of a streamed assistant reply.

ChatMessage rendered the reasoning block through two mutually exclusive branches gated on `message.parts.length`. When a `step-start` part arrived before the first `reasoning` part, the parts branch took over while `orderedParts` still filtered `step-start` out, so it rendered nothing and the fallback was already gone. The whole block blinked out until the first reasoning delta arrived.

The reasoning parts now render through one chronological list gated on renderable content, so a step-start-only window keeps the connecting indicator mounted. The leading reasoning block and its connecting placeholder share a stable each-key while the accordion open-state key keeps the real part key, so the same component instance survives the connecting to thinking transition without a remount or shimmer restart, and the auto-open logic in `reasoning-accordion-sync.ts` is untouched. `ReasoningTrigger` drives the connecting and thinking states from one shimmer instance so the animation no longer resets on the label change. A new pure helper `deriveOrderedParts` carries the ordering logic.

`bun scripts/static-checks.ts` passes on the changed files and `bun run test:unit` passes, including a new ordered-parts regression test that asserts a step-start-only parts array yields nothing to render.

### DIFF
--- a/src/lib/chat/ui/ChatMessage.svelte
+++ b/src/lib/chat/ui/ChatMessage.svelte
@@ -2,13 +2,12 @@
 	import { Message } from '$lib/components/prompt-kit/message';
 	import { Response } from '$lib/components/ai-elements/response';
 	import { ToolComposed } from '$lib/components/prompt-kit/tool';
-	import type { ToolPart } from '$lib/components/prompt-kit/tool/types.js';
 	import ChatAttachments from './ChatAttachments.svelte';
 	import ChatReasoning from './ChatReasoning.svelte';
 	import MessageBubble from './MessageBubble.svelte';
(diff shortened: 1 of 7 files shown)


stickerdaniel:
* **#444** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/444?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/444?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #445, opened 2026-06-01
feat(dev): deterministic per-project vite ports

### LINKED ISSUE #331: Look into portless
https://github.com/vercel-labs/portless

stickerdaniel:
Closes #331

Running several projects or worktrees from this template at once was painful: the dev and E2E vite servers were pinned to 5173/5174, so siblings collided, and a plain `vite dev` would silently increment into the pinned E2E port and break a parallel test run.

Both frontend ports are now derived per project from a hash of the working directory in `scripts/dev-ports.ts`, the single source consumed by the dev and dev:test wrappers, the Playwright config, and the e2e site-url resolver. Each fork and worktree lands on its own stable, bookmarkable port in two non-overlapping ranges (dev 20000-20999, test 21000-21999), so they never clash and the dev server can no longer creep into the test port. `--strictPort` stays, now with a friendly preflight that names the port and the lsof command on a collision, and `DEV_VITE_PORT`/`TEST_VITE_PORT` are escape hatches that fail loudly on an invalid value instead of silently falling back. The Convex backend ports were already dynamic, so they are untouched.

The harness is also portless-ready for anyone who prefers named URLs: set `PORTLESS_SITE_URL` to a vercel-labs/portless `.localhost` URL and the wrappers yield port control while the same origin flows into the Convex SITE_URL/trusted origin and Playwright's base URL. This stays optional, a plain clone needs nothing extra and never has to install portless. The README documents the `--no-tls` setup.

A vitest guard covers range membership, determinism, and override validation.

`bun run test:unit` passes (new dev-ports guard, 11 assertions). `bun scripts/static-checks.ts` is green including the project type-check. A full `bun run test:e2e` ran against the computed port (vite on :21184, backend on a random port) with signin/admin setup and the admin-users-table suite green, confirming the SITE_URL/Origin/trustedOrigin chain holds; the only residual failures were the known axe "execution context destroyed" dev-server flakes, unrelated to ports.

### DIFF
--- a/.env.convex.example
+++ b/.env.convex.example
@@ -76,9 +76,11 @@ AUTH_GITHUB_SECRET=
 # Base URL for OAuth redirects and deep links (can be preview URL)
 # Auto-derived from VERCEL_URL for preview deployments
 # Note: For email images, use EMAIL_ASSET_URL (always production)
-# Local dev default: http://localhost:5173 (matches `bun run dev`)
+# Local dev: auto-derived from the running Vite port (per-project, see
+# scripts/dev-ports.ts). Leave SITE_URL unset locally so a copied value can't pin
+# the trusted origin to a wrong port and break the admin sign-in.
 # For cloud/prod, set via: bunx convex env set SITE_URL https://yourapp.com
-SITE_URL=http://localhost:5173
+# SITE_URL=https://yourapp.com
 
 # ====================================
 # E2E Testing (REQUIRED for test setup)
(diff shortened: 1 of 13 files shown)


stickerdaniel:
* **#445** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/445?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/445?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #449, opened 2026-06-03
refactor(chat): runed listeners for drag-and-drop

### LINKED ISSUE #389: PromptInput drag-and-drop: addEventListener pairs should be useEventListener
## Problem

\`src/lib/components/prompt-kit/prompt-input/PromptInput.svelte\` (and the ai-elements twin) registers drag/drop handlers via raw \`addEventListener\` + \`removeEventListener\` pairs inside \`watch()\` cleanups:

\`\`\`ts
formRef.addEventListener('dragover', onDragOver);
formRef.addEventListener('drop', onDrop);
document.addEventListener('dragover', onDocumentDragOver);
document.addEventListener('drop', onDocumentDrop);
// ... cleanup elsewhere with removeEventListener
\`\`\`

Runed's \`useEventListener\` handles bound-target lifecycle automatically and avoids the asymmetry where one of the four removes could be missed under a refactor.

## Suggested fix

\`\`\`ts
import { useEventListener } from 'runed';

useEventListener(() => formRef, 'dragover', onDragOver);
useEventListener(() => formRef, 'drop', onDrop);
useEventListener(() => document, 'dragover', onDocumentDragOver);
useEventListener(() => document, 'drop', onDocumentDrop);
\`\`\`

Cleanup happens on unmount + when formRef ref reassigns.

## See also

- #384, #385, #386, #387, #388 — runed-consistency theme

stickerdaniel:
Closes #389

The prompt-input registered its drag-and-drop handlers through two hand-rolled watch blocks that called addEventListener and removeEventListener directly, with the Files-only dragover and drop handlers duplicated across both blocks. The asymmetry between the four add and four remove calls is the kind of thing a later refactor can quietly break, and the listeners do not auto-dispose against the component lifecycle the way the rest of the codebase now does.

This swaps both blocks for runed's useEventListener. The handlers are defined once and bound via reactive target getters: `() => formRef` for the form scope and `() => (globalDrop ? document : null)` for the global listeners, so the global handlers attach only while globalDrop is set and detach as soon as it clears. Cleanup runs automatically on unmount and when a target getter changes, matching the existing useMutationObserver usage elsewhere. The third watch block (file-input sync) is untouched.

`bun scripts/static-checks.ts` passes and the Svelte autofixer reports no issues on the changed file.

### DIFF
--- a/src/lib/components/ai-elements/prompt-input/PromptInput.svelte
+++ b/src/lib/components/ai-elements/prompt-input/PromptInput.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
 	import { cn } from '$lib/utils';
 	import type { Snippet } from 'svelte';
-	import { watch } from 'runed';
+	import { useEventListener, watch } from 'runed';
 	import { onMount } from 'svelte';
 	import {
 		AttachmentsContext,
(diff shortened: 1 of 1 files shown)


stickerdaniel:
* **#449** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/449?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/449?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

stickerdaniel:
Triage of Qodo's findings: the "unguarded document reference" bug is a false positive. `useEventListener` reads the target getter only inside its own `$effect`, which never runs during SSR (same as the previous `watch`-based code), so `document` is touched client-side only after hydration. No change needed.

PR #450, opened 2026-06-03
refactor(hooks): replace manual timers/listeners with runed

### LINKED ISSUE #385: Runed adoption: use-clipboard, use-haptic, use-auto-scroll still roll their own listener/timer plumbing
## Problem

Three shared hooks under `src/lib/hooks/` carry manual `setTimeout` + `addEventListener` + `MutationObserver` patterns that runed already covers cleanly. The current code works but doesn't auto-dispose against component lifecycle, which the AGENTS.md \"Before creating our own utilities, research the runed library\" rule asks us to use.

### 1. `src/lib/hooks/use-clipboard.svelte.ts`
- Lines 31, 48-58: manual `private timeout` + `setTimeout` + `clearTimeout` to reset `#copiedStatus` after delay.
- Runed equivalent: `useDebounce(fn, () => this.delay)` + `cancel()` on re-copy. Auto-disposes.

### 2. `src/lib/hooks/use-haptic.svelte.ts`
- Lines 21, 57, 61-63, 70: `#activeTimeout` + `setTimeout` / `clearTimeout` cascade for the `isActive` flag.
- Same `useDebounce` treatment with reactive wait (`() => this.#activeDuration`).

### 3. `src/lib/hooks/use-auto-scroll.svelte.ts`
- Lines 50, 55, 58, 67-69: `el.addEventListener('scroll', ...)`, `window.addEventListener('resize', ...)`, `new MutationObserver(...)` with manual `removeEventListener` + `observer.disconnect()` in the cleanup function.
- Runed equivalents: `useEventListener(() => el, 'scroll', cb)`, `useEventListener(() => window, 'resize', cb)`, `useMutationObserver(() => el, cb, { childList: true, subtree: true })`. Each auto-disposes on component unmount.

## Note on origin

`use-clipboard` and `use-auto-scroll` are installed from `@ieedan/shadcn-svelte-extras`. The cleanest upstream fix lives in that package, but until those land the saas-starter's vendored copies could pre-empt them so consumers don't inherit the drift.

## Suggested fix

Working diff verified in a downstream fork — happy to send a PR. Net change per file ~10-20 lines removed, no behaviour change.

## See also

- #384 (RouteProgress timer leak — same theme)

stickerdaniel:
Closes #385

Three shared hooks carried hand-rolled setTimeout, addEventListener, and MutationObserver plumbing that runed already wraps with lifecycle-aware cleanup. The vendored use-clipboard and use-auto-scroll come from @ieedan/shadcn-svelte-extras; use-haptic is local to this repo.

use-haptic and use-clipboard move their reset timers to useDebounce. In use-haptic the per-trigger duration flows through a reactive `#lastDuration` field, since useDebounce caches its wait via a `$derived` and a plain field would never recompute. Both schedule calls swallow the cancel rejection that useDebounce raises on a pending promise. use-clipboard keeps its pre-reset on rapid re-copy via the `pending` flag, so the status transition still re-fires.

use-auto-scroll adopts the upstream fix from ieedan/shadcn-svelte-extras#391. The scroll and resize listeners become useEventListener and the MutationObserver becomes useMutationObserver, all auto-disposing on unmount. The initial positioning moves into its own untracked effect so it runs when the ref binds rather than re-running on every scroll, a small intentional improvement over the previous behaviour.

`bun scripts/static-checks.ts` passes for all three files and the Svelte autofixer reports no issues.

### DIFF
--- a/src/lib/hooks/use-auto-scroll.svelte.ts
+++ b/src/lib/hooks/use-auto-scroll.svelte.ts
@@ -2,6 +2,9 @@
 	Installed from @ieedan/shadcn-svelte-extras
 */
 
+import { untrack } from 'svelte';
+import { useEventListener, useMutationObserver } from 'runed';
+
 /** Use this on a vertically scrollable container to ensure that it automatically scrolls to the bottom of the content.
  *
  * ## Usage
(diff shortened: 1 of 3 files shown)


stickerdaniel:
* **#450** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/450?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/450?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

stickerdaniel:
Triage of Qodo's findings: the "clipboard debounce delay undefined" bug is technically a false positive (the lazy `$derived` wait reads `delay` at the first `copy()`, after the constructor has run, and `delay` is static). Moved it to a reactive `#delay = $state(500)` in 24c37e2 anyway, matching use-haptic and dropping the field-init-order dependency.

PR #478, opened 2026-06-04
fix(team-switcher): use cmdOrCtrl shortcut label

### LINKED ISSUE #476: team-switcher hardcodes ⌘ in the shortcut display
team-switcher.svelte:55 renders a hardcoded Command glyph, so Windows/Linux users see a key they cannot press. Every other shortcut display goes through `cmdOrCtrl` from $lib/hooks/is-mac.svelte (sidebar configs, command trigger).

proposed fix:

```diff
-<DropdownMenu.Shortcut>⌘{index + 1}</DropdownMenu.Shortcut>
+<DropdownMenu.Shortcut>{cmdOrCtrl}{index + 1}</DropdownMenu.Shortcut>
```

with the corresponding import.

ded-furby:
## Summary
- replace the hardcoded `⌘` badge in the team switcher with the shared `cmdOrCtrl` helper
- align the component with the repo rule to avoid platform-specific shortcut text

Closes #476

## Test URLs
- Not applicable: this is a static shortcut label change inside the existing sidebar team switcher component, and I did not run the full app locally in this environment.

## Screenshot
- Not applicable for this small text-only shortcut badge update.

## Validation
- `bun scripts/static-checks.ts src/lib/components/team-switcher.svelte`

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -13,6 +13,7 @@ import requireMarketingRouteRegistrationRule from './eslint/rules/require-market
 import noHardcodedAriaLabelRule from './eslint/rules/no-hardcoded-aria-label.js';
 import noHardcodedSrOnlyRule from './eslint/rules/no-hardcoded-sr-only.js';
 import noDebounceInRuneRule from './eslint/rules/no-debounce-in-rune.js';
+import noHardcodedModifierKeysRule from './eslint/rules/no-hardcoded-modifier-keys.js';
 
 const gitignorePath = path.resolve(import.meta.dirname, '.gitignore');
 const localPlugin = {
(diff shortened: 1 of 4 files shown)


stickerdaniel:
Reverted 7d4e378. The shortcut renders inside `DropdownMenu.Content`, which bits-ui only mounts while the menu is open, so it never appears in SSR HTML and a hydration mismatch cannot occur. The Qodo warning is a false positive here. Inline `{cmdOrCtrl}` also matches how `command-trigger.svelte` uses the helper. The cmdOrCtrl fix itself looks good, thanks!

stickerdaniel:
Thanks!

PR #504, opened 2026-06-06
fix(i18n): Translate hardcoded UI strings and validation errors

### LINKED ISSUE #467: Hardcoded UI strings bypass Tolgee; soft_delete key missing in de/es/fr
Hardcoded user-facing strings that bypass Tolgee, each with a localized sibling already in place:

- [ ] signin/SignUpForm.svelte:104: placeholder "Your name"; `settings.account.name.placeholder` already exists in all four locales and is used by account settings
- [ ] ui/confirm-delete-dialog/confirm-delete-dialog.svelte:118 and 138: hardcoded placeholder and English button fallback texts
- [ ] community-chat/+page.svelte:150: hardcoded "You" author fallback
- [ ] `admin.resources.projects.actions.soft_delete` exists only in en.json, missing in de/es/fr. The whole `admin.resources.*` namespace looks unreferenced in src/, so check whether it is dead before translating.

The `m@example.com` placeholders in the auth forms are language-neutral and can stay.

Email template strings are tracked separately in #378.

proposed fix: bind to the existing keys or add keys to all four locale files, then `bun run i18n:push`.

### LINKED ISSUE #468: Admin add-email form renders raw English validation errors end-to-end
The admin add-email remote form, the reference example for `form(schema, handler)`, renders raw English at every layer:

- email-schema.ts:9 hardcodes "Please enter a valid email address" while every other email schema emits the `validation.email.invalid` key
- add-email-dialog.svelte:73 passes raw `issues()` to `Field.Error` without the shared `translateValidationErrors` helper
- data.remote.ts:27 returns `invalid(issue.email('This email already exists'))` as a literal English string

Related: src/lib/schemas/auth.ts:4 calls `v.email()` with no message argument, so Valibot's built-in English default leaks the same way. And founder-welcome-card.svelte:40 validates the reply-to address server-side only (ConvexError surfaces as a generic toast), with no client-side schema, unlike email-settings.

proposed fix: emit Tolgee keys from both schema and server handler, translate at render via `translateValidationErrors`, and add a client-side reply-to check reusing the shared email pipe.

### LINKED ISSUE #470: ConvexCursorTableShell hardcodes admin.users.* keys; recipients table shows "Search users..."
convex-cursor-table-shell.svelte:141-195 hardcodes `admin.users.*` keys for the page indicator and all pagination buttons, although the shell is the reusable table chrome. The second consumer inherits the wrong namespace: notification-recipients-table.svelte:258 and 271 borrow `admin.users.search_placeholder` and `admin.users.rows_per_page`, so the recipients table literally shows "Search users..." in every locale even though `admin.resources.notification_recipients.search_placeholder` already exists.

proposed fix: introduce a generic `table.*` namespace inside the shell (or accept label props/snippets like `searchPlaceholder` already does), and point the recipients table at its own keys.

stickerdaniel:
Closes #467
Closes #468
Closes #470

Several surfaces bypassed Tolgee and rendered raw English in every locale. The reusable cursor table shell hardcoded `admin.users.*` keys plus an English rows-per-page prop default, so the notification recipients table showed "Search users..." and borrowed user-table labels. The signup name placeholder, the confirm-delete dialog texts and the optimistic chat author fallback were hardcoded. The admin add-email form showed untranslated validation errors at every layer, and an invalid founder welcome reply-to only failed server-side as a generic toast.

The table shell now uses a generic `table.*` namespace with a translated render-time fallback for the rows-per-page label, and the recipients table has its own search placeholder. The remaining hardcoded strings route through existing or new `common.*` keys. The add-email schema and handler return translation keys that a new `translateRemoteFormIssues` helper resolves through Tolgee at render time, following the existing `translateValidationErrors` pattern. The founder welcome reply-to field is validated client-side with the shared email schema, shows a translated inline error and disables save while invalid; empty stays valid. New keys exist in all four locales and were pushed to Tolgee tagged draft; the six superseded `admin.users.*` keys are removed locally and will be cleaned up remotely after merge.

The soft-deleted users point from #467 was already resolved by #503.

`bun run test:unit` and `bun scripts/static-checks.ts` pass.

### DIFF
--- a/src/i18n/de.json
+++ b/src/i18n/de.json
@@ -131,6 +131,7 @@
 			"preference_enabling": "{field} wird aktiviert...",
 			"preference_update_failed": "Einstellung konnte nicht aktualisiert werden",
 			"recipient_count": "{count, plural, other {# Empfänger}}",
+			"recipients_search_placeholder": "Benachrichtigungsempfänger durchsuchen...",
 			"save": "Einstellungen speichern",
 			"saved": "Einstellungen gespeichert",
 			"saving": "Speichern...",
@@ -255,20 +256,12 @@
 			"menu_open": "Menü öffnen",
 			"name": "Name",
 			"no_results": "Keine Ergebnisse.",
-			"page_indicator": "Seite {current} von {total}",
-			"pagination": {
-				"first": "Zur ersten Seite",
-				"last": "Zur letzten Seite",
-				"next": "Zur nächsten Seite",
-				"previous": "Zur vorherigen Seite"
-			},
 			"provider": "Anbieter",
 			"provider_email": "E-Mail",
 			"provider_github": "GitHub",
 			"provider_google": "Google",
 			"provider_passkey": "Passkey",
 			"role": "Rolle",
-			"rows_per_page": "Zeilen pro Seite",
 			"search_placeholder": "Benutzer suchen...",
 			"select_all": "Alle auswählen",
 			"select_row": "Zeile auswählen",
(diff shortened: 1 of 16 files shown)


stickerdaniel:
* **#504** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/504?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/504?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #506, opened 2026-06-08
fix(pricing): Sync free-tier copy with config and preserve checkout locale

### LINKED ISSUE #455: Pricing copy disagrees with autumn.config.ts limits (10 vs 3 free messages, 200 vs 500 comment)
The configured limits and the user-facing copy disagree:

- autumn.config.ts grants the free tier `included: 3` messages/month; the pricing page says "10 messages per month" in all four locales (en.json key `pricing...free[0]`, same in de/es/fr), in pricing/page.md.ts:19, and in README.md:345.
- the Pro docstring in autumn.config.ts:45 says "200 AI chat messages/month" while the item grants `included: 500`.

Side note: `ai_chat.pro_required` exists in all locale files but is referenced nowhere.

proposed fix: decide the canonical numbers, then sync the config comment, the four locale files, page.md.ts and the README in one pass; remove the dead key.

### LINKED ISSUE #494: Checkout successUrl hardcodes a lang-less /app path, dropping the user's locale on return
Two checkout call sites build the Autumn `successUrl` without the `[[lang]]` prefix, so the post-checkout return loses the user's chosen locale.

- `nav-user.svelte:56`: the user-menu upgrade handler `handleUpgrade` sends `successUrl: page.url.origin + '/app/community-chat?upgraded=true'`.
- `pricing-three.svelte:75`: `handleCheckout` sends the identical lang-less `successUrl`, even though the same function's pre-signin branch (`pricing-three.svelte:63-69`) already preserves locale via `localizedHref('/signin')` plus `page.url.pathname`.

After checkout, Autumn/Stripe redirects back to that prefix-less URL. `shouldBypassLanguageRedirect` (`hooks.server.ts:81`) only exempts `/api`, `/llms.txt`, `/robots.txt`, and `/sitemap.xml`, so `/app/community-chat` falls through to the `^\/([a-z]{2})(\/|$)` prefix check (`hooks.server.ts:153`), and `handleLanguage` 307-redirects it using the `accept-language` header (`hooks.server.ts:159`, redirect at `hooks.server.ts:177`). A user whose UI language differs from their browser language lands back on the wrong locale, with an extra redirect hop. It self-corrects on the next navigation through `localizedHref`, so this is cosmetic, but forks copying the pricing block inherit the same behavior.

The in-page checkout buttons already do this right by building the URL from the current page: `community-chat/+page.svelte:169` and `ai-chat/+page.svelte:66` both use `new URL(page.url.href)` and set `upgraded=true` on it, which keeps the active locale prefix.

proposed fix: in both call sites, build `successUrl` from the current page so the locale prefix survives the round-trip:

```ts
const successUrl = new URL(page.url.href);
successUrl.searchParams.set('upgraded', 'true');
// ...
successUrl: successUrl.href
```

stickerdaniel:
Closes #455
Closes #494

The pricing surfaces and the billing config disagreed, and two checkout call sites dropped the user's locale on return. The free tier copy advertised "10 messages per month" in all four locales, the agent-facing pricing markdown and the README, while `autumn.config.ts` grants only 3. The Pro docstring claimed "200 AI chat messages" while the item grants 500, and an orphaned `ai_chat.pro_required` key (claiming 30) was rendered nowhere. Separately, `nav-user.svelte` and the pricing block built the Autumn `successUrl` by concatenating the origin with a hardcoded `/app/community-chat`, so the post-checkout redirect lost the `[[lang]]` prefix and bounced through a 307 based on the browser's accept-language header.

The config is the enforced source of truth, so the free-tier count now reads 3 everywhere and the Pro docstring reads 500. The dead `ai_chat.pro_required` key and a parallel set of unused flat `free[n]`/`pro[n]`/`enterprise[n]` pricing keys (the component renders the nested tier objects via `getFeatureKeys`, never the flat ones) are removed from all four locale files. Both checkout call sites now build the success URL through `localizedHref('/app/community-chat')`, keeping the active locale and the existing community-chat destination, matching the in-page chat upgrade flows. The locale files are pushed to Tolgee after merge.

`bun run test:unit` (including locale parity) and `bun scripts/static-checks.ts` pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -348,7 +348,7 @@ Sign in with email/password, Google, GitHub, or passkeys. Powered by [Better Aut
 
 ### Billing
 
-[Autumn](https://docs.useautumn.com/welcome) sits on top of Stripe and lets you define pricing tiers and usage gates without writing webhook handlers. Configure your products in `autumn.config.ts` or generate a starter config at [app.useautumn.com/sandbox/quickstart](https://app.useautumn.com/sandbox/quickstart). Ships with a Free tier (10 messages/month) and a Pro tier ($10/month, unlimited). The community chat enforces quotas, warns users when they are running low, and offers an upgrade flow.
+[Autumn](https://docs.useautumn.com/welcome) sits on top of Stripe and lets you define pricing tiers and usage gates without writing webhook handlers. Configure your products in `autumn.config.ts` or generate a starter config at [app.useautumn.com/sandbox/quickstart](https://app.useautumn.com/sandbox/quickstart). Ships with a Free tier (3 messages/month) and a Pro tier ($10/month, unlimited). The community chat enforces quotas, warns users when they are running low, and offers an upgrade flow.
 
 ### Admin Panel
 
--- a/autumn.config.ts
+++ b/autumn.config.ts
@@ -43,7 +43,7 @@ export const free = plan({
 });
 
 /**
- * Pro tier with unlimited community chat and 200 AI chat messages/month.
+ * Pro tier with unlimited community chat and 500 AI chat messages/month.
  */
 export const pro = plan({
 	id: 'pro',
--- a/src/blocks/pricing/pricing-three.svelte
+++ b/src/blocks/pricing/pricing-three.svelte
@@ -70,9 +70,11 @@
 		}
 
 		// Proceed with checkout for authenticated users
+		const successUrl = new URL(localizedHref('/app/community-chat'), page.url.origin);
+		successUrl.searchParams.set('upgraded', 'true');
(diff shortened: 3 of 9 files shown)


stickerdaniel:
* **#506** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/506?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/506?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #507, opened 2026-06-08
fix(convex): Rate-limit chat mutations and enforce community quota server-side

### LINKED ISSUE #451: Community chat: free-tier message limit is enforced client-side only
The `send` mutation in src/lib/convex/messages.ts:63 only validates body length, inserts the message, then schedules usage tracking fire-and-forget. The only quota gate is the client-side `hasMessagesAvailable` check in community-chat, and the doc comment above the mutation states billing is enforced client-side. Calling `api.messages.send` directly bypasses the free-tier message limit entirely.

proposed fix: check entitlement before the insert, mirroring `createAIResponse`:

```ts
const sdk = await getAutumnSdk();
const { data } = await sdk.check({ customer_id: ctx.user._id, feature_id: 'messages' });
if (!data?.allowed) throw new ConvexError('Message limit reached');
```

Keep the client guard for UX only.

### LINKED ISSUE #495: Community chat send mutation has no rate limiting
`messages.send` (`src/lib/convex/messages.ts:63-81`) is an `authedMutation` with only a 2000-char length cap and no rate limiter, unlike its two sibling send mutations. `aiChat/messages.ts` `sendMessage` calls `aiChatRateLimiter.limit()` at `aiChat/messages.ts:56`, and `support/messages.ts` `sendMessage` calls `supportRateLimiter.limit()` at `support/messages.ts:49`. The client `handleSend` in `community-chat/+page.svelte:127` only gates on a billing balance (`hasMessagesAvailable`) and an `isSending` flag, so a direct `api.messages.send` call floods the shared community channel and fires the per-call `trackMessageUsage` Autumn action (`messages.ts:75`) at unbounded frequency.

The same gap shows up in four more mutations that have rate-limited siblings:

- `storage.ts:16-27` profile-image `generateUploadUrl` mints storage upload URLs with no `rateLimiter.limit()`, while the near-identical `aiChat/files.ts:27` and `support/files.ts:39` rate-limit theirs.
- `storage.ts:43-89` `updateProfileImage` writes and deletes `_storage` blobs on every call with no rate limiting.
- `support/threads.ts:139-160` `createThread` is anonymous-reachable and creates an agent thread plus a `supportThreads` row per call with no `rateLimiter.limit()`, while `support/messages.ts:24` rate-limits the same callers.
- `aiChat/threads.ts:60-78` `createThread` creates an agent thread plus an `aiChatThreads` row per call with no rate limiting, while `aiChat/messages.ts:25` does.

Related to #451, which covers billing enforcement on the same `messages.send` mutation.

proposed fix: add a `messages` rate limiter keyed by `ctx.user._id` and call `.limit()` at the top of `messages.send` the way the support and ai-chat send mutations do, then extend the same `rateLimiter.limit()` guard to the two `storage.ts` mutations and both `createThread` mutations so the upload and thread-creation paths match their rate-limited message siblings.

stickerdaniel:
Closes #451
Closes #495

The community chat send mutation enforced the free-tier message quota only on the client and, unlike its AI chat and support siblings, had no rate limiter, so a direct `api.messages.send` call could flood the shared channel and bypass the limit entirely. The same missing-rate-limiter gap existed in five more mutations: both profile image storage mutations, and `createThread` plus `getOrCreateWarmThread` in both the support and AI chat features, where the warm-thread variant would otherwise sidestep a createThread-only limit.

Every one of these mutations now calls a per-user rate limiter at the top, keyed on the authenticated user (a shared global bucket for spoofable anonymous support callers, sized for the marketing-page chatbar that creates a warm thread on a new visitor's first keystroke). Thread-creation limits only consume a token on the actual creation branch so idempotent warm-thread reads stay free. The Autumn entitlement check needs an action context, so community chat quota enforcement runs in the already-scheduled follow-up rather than the mutation: it removes an over-limit message and tracks usage only when the send is allowed. That backstop is reached by direct API calls and by legitimate sends right at the quota boundary, where the message briefly appears and is then removed; a failed Autumn check never deletes a legitimate message, and the decision logic is unit-tested.

All rate-limit errors are structured with a retryAfter so clients can react: community chat and the avatar settings show translated toasts with the retry time and restore the user's input, the warm-thread auto-creation effects in the app layout and the ai-chat page back off instead of retrying at network pace, and the support chatbar clears its cached creation promise on failure so one rejected thread creation no longer disables it until reload.

`bun run check:convex`, `bun run test:unit` (including locale parity and the new backstop tests) and `bun scripts/static-checks.ts` pass.

### DIFF
--- a/src/i18n/de.json
+++ b/src/i18n/de.json
@@ -605,6 +605,7 @@
 			"upgrade_tooltip": "Auf Pro upgraden für unbegrenzte Nachrichten"
 		},
 		"messages": {
+			"rate_limited": "Sie senden zu schnell Nachrichten. Versuchen Sie es in {seconds}s erneut.",
 			"send_failed": "Nachricht konnte nicht gesendet werden",
 			"sent_success": "Nachricht gesendet!"
 		},
--- a/src/i18n/de.json
+++ b/src/i18n/de.json
@@ -928,6 +928,7 @@
 		"account": {
 			"avatar": {
 				"alt": "Profilvorschau",
+				"rate_limited": "Zu viele Upload-Versuche. Versuchen Sie es in {seconds}s erneut.",
 				"ready": "Bild bereit zum Speichern",
 				"removed": "Bild entfernt - klicken Sie auf Speichern zur Bestätigung",
 				"select_error": "Bitte wählen Sie eine Bilddatei aus",
--- a/src/i18n/en.json
+++ b/src/i18n/en.json
@@ -928,6 +928,7 @@
 		"account": {
 			"avatar": {
 				"alt": "Profile preview",
+				"rate_limited": "Too many upload attempts. Try again in {seconds}s.",
 				"ready": "Image ready to save",
 				"removed": "Image removed - click Save to confirm",
 				"select_error": "Please select an image file",
--- a/src/i18n/en.json
+++ b/src/i18n/en.json
@@ -605,6 +605,7 @@
 			"upgrade_tooltip": "Upgrade to Pro for unlimited messages"
 		},
 		"messages": {
+			"rate_limited": "You're sending messages too fast. Try again in {seconds}s.",
 			"send_failed": "Failed to send message",
 			"sent_success": "Message sent!"
 		},
--- a/src/i18n/es.json
+++ b/src/i18n/es.json
@@ -605,6 +605,7 @@
 			"upgrade_tooltip": "Actualiza a Pro para mensajes ilimitados"
 		},
 		"messages": {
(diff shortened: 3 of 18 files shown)


stickerdaniel:
* **#507** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/507?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/507?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #509, opened 2026-06-10
fix(convex): Count Autumn usage atomically with the entitlement check

stickerdaniel:
The community chat quota backstop and the AI chat response action both enforced metered features with a read-only Autumn check followed by a separate track call. Between the two HTTP calls every concurrent request reads the same stale balance, so with one unit left, N parallel sends all pass and overshoot the limit. For AI chat the window spans the entire multi-second stream.

Autumn closes this server-side: a check with send_event deducts the usage and computes allowed in one atomic operation with overage behavior reject, so concurrent calls serialize and exactly the remaining quota passes. The new checkAndCountUsage helper in autumn.ts wraps this and documents the semantics once: counted (usage recorded), denied (nothing deducted), or unavailable (Autumn errored, nothing deducted, callers fail open). refundUsage credits a unit back via a negative track value, which Autumn documents for exactly this purpose.

The community backstop now removes a message only on a definitive denied outcome. createAIResponse counts the unit before generating and refunds it when generation fails, so a failed response never costs a message. The refund window ends once the stream is consumed, so a delivered response is never refunded by later denormalization errors. A thrown check now also fails open for AI chat and generation proceeds uncounted instead of the action dying before the stream.

The remaining Autumn call sites were audited: the title generation check stays intentionally check-only (its usage unit is tracked by the response path) and the client-side useCustomer gates are advisory UI state.

See also: #451, #507

`bun run check:convex`, `bun run test:unit` and `bun scripts/static-checks.ts` pass.

### DIFF
--- a/src/lib/convex/__tests__/autumn.test.ts
+++ b/src/lib/convex/__tests__/autumn.test.ts
@@ -1,0 +1,127 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+vi.mock('../env', () => ({
+	requireEnv: vi.fn(() => 'sk_test')
+}));
+
+vi.mock('../auth', () => ({
+	authComponent: {
+		getAuthUser: vi.fn()
+	}
+}));
+
+vi.mock('../_generated/api', () => ({
+	components: { autumn: {} },
+	internal: {}
+}));
+
+vi.mock('@useautumn/convex', () => ({
+	Autumn: class {
+		api() {
+			return {};
+		}
+	}
+}));
+
+const check = vi.fn();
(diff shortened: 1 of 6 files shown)


stickerdaniel:
* **#509** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/509?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/509?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #512, opened 2026-06-11
fix(ci): Pass CF check-run summary via env block

### LINKED ISSUE #481: e2e-preview-cf.yml interpolates the CF check-run summary into bash, a shell-injection sink
In `.github/workflows/e2e-preview-cf.yml:54` the "Extract preview URL" step assigns `SUMMARY="${{ github.event.check_run.output.summary }}"`. GitHub Actions substitutes the expression into the script text before bash parses it, so the Cloudflare check-run summary becomes part of the shell source. The summary embeds branch and deployment data that anyone who can push a branch controls, and a double quote or `$(...)` in that text breaks out of the assignment and runs arbitrary commands.

This runs inside the `e2e-tests` job, which holds `statuses: write` (`e2e-preview-cf.yml:33`) and exposes `AUTH_E2E_TEST_SECRET`, `CF_ACCESS_CLIENT_ID`, and `CF_ACCESS_CLIENT_SECRET` as env vars (`e2e-preview-cf.yml:42`-`44`), all reachable from the injected script. The `check_run` trigger runs with base-repo secrets regardless of fork origin, so a fork PR reaches this sink too. The other `github.event.*` uses in the file sit in `if:` and `with: ref:`, which are not shell sinks.

proposed fix: pass the value through `env:` and reference `"$SUMMARY"` so the summary never enters the script body:

```yaml
      - name: Extract preview URL
        env:
          SUMMARY: ${{ github.event.check_run.output.summary }}
        run: |
          PREVIEW_URL=$(echo "$SUMMARY" | grep 'Preview Alias URL:' | sed 's/.*Preview Alias URL: //' | tr -d '[:space:]')
```

stickerdaniel:
Closes #481

The `Extract preview URL` step in `.github/workflows/e2e-preview-cf.yml` assigned `SUMMARY="${{ github.event.check_run.output.summary }}"` inline, so GitHub Actions substituted the Cloudflare check-run summary into the script text before bash parsed it. The summary embeds branch and deployment data that anyone who can push a branch controls, and a double quote or `$(...)` in it would break out of the assignment and run arbitrary commands in a job that holds `statuses: write` and exposes `AUTH_E2E_TEST_SECRET`, `CF_ACCESS_CLIENT_ID`, and `CF_ACCESS_CLIENT_SECRET`. The `check_run` trigger runs with base-repo secrets regardless of fork origin, so fork PRs reach this sink too.

The summary is now passed to the step through an `env:` block and the script reads the quoted `"$SUMMARY"` variable, so the value never enters the parsed script body and expansion is injection-safe. The remaining `run:` interpolations in the file are `github.repository`, `check_run.head_sha`, `job.status`, and `github.run_id`, all GitHub-generated values (a GitHub-generated hex SHA), and the other `github.event` uses sit in `if:` conditions and a `with: ref:` input, none of which are shell sinks.

`bun scripts/static-checks.ts .github/workflows/e2e-preview-cf.yml` passes, and `actionlint` reports no new findings beyond the pre-existing runner-label and SC2086/SC2129 notes already present on `main`. The extraction itself runs unchanged on the next CF preview build that triggers this workflow.

### DIFF
--- a/.github/workflows/e2e-preview-cf.yml
+++ b/.github/workflows/e2e-preview-cf.yml
@@ -50,8 +50,9 @@ jobs:
           ref: ${{ github.event.check_run.head_sha }}
 
       - name: Extract preview URL
+        env:
+          SUMMARY: ${{ github.event.check_run.output.summary }}
         run: |
-          SUMMARY="${{ github.event.check_run.output.summary }}"
           PREVIEW_URL=$(echo "$SUMMARY" | grep 'Preview Alias URL:' | sed 's/.*Preview Alias URL: //' | tr -d '[:space:]')
           if [ -z "$PREVIEW_URL" ]; then
             echo "::error::No Preview Alias URL found in check_run output"


stickerdaniel:
* **#512** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/512?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/512?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #514, opened 2026-06-11
chore(scripts): Use spawn args for convex env list

### LINKED ISSUE #473: validate-convex-env builds a shell string; build:emails is the only script on tsx
Two deviations from the cross-platform script conventions:

- scripts/validate-convex-env.ts:79 builds a shell command via string interpolation and `execSync`, so quoting of `--deployment-name` values depends on the host shell. Every other script passes array args without a shell.
- package.json `build:emails` is the only script that runs under tsx; build-emails.ts uses no Bun-specific APIs, and tsx stays a devDependency for this one script.

proposed fix: switch to `spawnSync('bunx', ['convex', 'env', 'list', ...flags])` (or reuse the deploy utils runner); run build-emails under bun and remove the tsx devDependency if nothing else needs it.

stickerdaniel:
Closes #473

`scripts/validate-convex-env.ts` built the `bunx convex env list` command as a single shell string with `execSync`, so quoting of `--deployment-name` values depended on the host shell, while every other script in the repo passes argument arrays without a shell. `build:emails` was also the only script still running under `tsx`, keeping a devDependency around for one script even though `scripts/build-emails.ts` needs nothing tsx-specific.

The validation script now builds the deployment flag as an argument array and runs it through `runCommandCapture` from `scripts/deploy/utils`, printing the captured stderr and exiting non-zero on failure. The CLI hint for missing variables derives its flag suffix from the same array. `build:emails` runs `scripts/build-emails.ts` under `bun` directly, and the `tsx` devDependency is removed via `bun remove tsx` since nothing else references it.

`bun scripts/static-checks.ts scripts/validate-convex-env.ts package.json` passes, `bun run build:emails` renders 6/6 templates under bun, `bun scripts/validate-convex-env.ts` succeeds against the dev deployment and exits 1 with the CLI error for an inaccessible `--deployment-name`, and `bun run test:unit` passes with 486 tests.


stickerdaniel:
* **#514** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/514?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/514?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #516, opened 2026-06-11
refactor(convex): Align schema userId conventions

### LINKED ISSUE #464: Schema consistency: adminProfiles by_userId naming, supportThreads.userId optionality
Two small schema inconsistencies that get copied as-is:

- adminProfiles names its index `by_userId` (schema.ts:127) while every other userId index in the schema is `by_user`. Two `withIndex('by_userId')` call sites in admin/founderWelcome.
- supportThreads.userId is optional (support/supportThreadFields.ts:11) while aiChatThreads.userId is required (schema.ts:180). Support code treats it as guaranteed (ownership checks, `?? thread.userId` workarounds), so the optionality is misleading.

proposed fix: rename the index to `by_user`; make supportThreads.userId required to match the actual write contract, or document the anonymous-orphan case if it is intended.

stickerdaniel:
Closes #464

The schema had two inconsistencies that get copied as-is by anyone extending the template. `adminProfiles` named its index `by_userId` while every other userId index in the schema is `by_user`, and `supportThreads.userId` was declared optional even though every write path (`createSupportThreadRecord`, the E2E test seeder, and the anonymous-ticket migration) always sets it to an authenticated or `anon_*` owner ID. The misleading optionality forced workarounds like the `?? thread.userId` fallback in `listThreads` and wider return validators than the actual data contract.

This renames the `adminProfiles` index to `by_user` and updates the three `withIndex` call sites in `admin/founderWelcome`. `supportThreads.userId` is now `v.string()` in `supportThreadFields.ts`, which also tightens the shared `vSupportMetadata` admin validator that spreads these fields. The dead `supportThread.userId ?? thread.userId` fallback in `listThreads` is simplified to `supportThread.userId`, and the return validators that surface this field (`listThreads`, `listThreadsForAdmin`, `getThreadForAdmin`) are narrowed to `v.string()`. `getThread` keeps an optional `userId` because it spreads the agent component thread, whose `userId` is optional upstream.

Index renames are code-only and need no migration. Every insert and migration path at HEAD sets `userId`, but between January and March 2026 `createThread` was a public mutation that accepted an optional `userId` and inserted it verbatim, so a production deployment from that window could still hold rows without it. Before merging into such a deployment, check the dashboard for `supportThreads` rows missing `userId` (the schema push fails validation if any exist) and backfill or delete them first.

Verification: `bun scripts/static-checks.ts` on the changed files, `bun run check:convex`, and `bun run test:unit` (486 tests) all pass.

### DIFF
--- a/src/lib/convex/admin/founderWelcome/mutations.ts
+++ b/src/lib/convex/admin/founderWelcome/mutations.ts
@@ -74,7 +74,7 @@ export const updateConfig = adminMutation({
 		// Upsert per-user profile (name, title, replyTo)
 		const existingProfile = await ctx.db
 			.query('adminProfiles')
-			.withIndex('by_userId', (q) => q.eq('userId', adminUserId))
+			.withIndex('by_user', (q) => q.eq('userId', adminUserId))
 			.unique();
 
 		if (existingProfile) {
--- a/src/lib/convex/admin/founderWelcome/queries.ts
+++ b/src/lib/convex/admin/founderWelcome/queries.ts
@@ -63,7 +63,7 @@ async function readFounderWelcomeConfig(ctx: QueryCtx): Promise<FounderWelcomeCo
 	// Read per-user profile for contact person
 	const profile = await ctx.db
 		.query('adminProfiles')
-		.withIndex('by_userId', (q) => q.eq('userId', contactUserId))
+		.withIndex('by_user', (q) => q.eq('userId', contactUserId))
 		.unique();
 
 	// Read global email config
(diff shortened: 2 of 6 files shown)


stickerdaniel:
* **#516** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/516?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/516?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #517, opened 2026-06-11
test(e2e): Replace conditional skips with auto-waits

### LINKED ISSUE #508: Flaky admin-settings e2e tests skip themselves instead of auto-waiting
## Problem

`e2e/admin-settings.spec.ts` has two tests that `test.skip()` themselves when a timing race makes a snapshot read empty, instead of testing anything:

- `filters recipients by type`: `const n = await allRows.count(); if (n < 1) { test.skip(); return; }`
- `toggles notification preference`: `if (!(await toggle.isVisible())) { test.skip(); return; }`

Both read a **non-retrying** snapshot (`locator.count()`, `locator.isVisible()`) right after the `recipients-loading` flag clears but before the reactive table has painted its rows. If the read lands in that window it sees nothing and the test skips. On a slow or loaded CI run these tests silently produce no signal.

A test that skips itself to dodge flakiness is a green-light hack, not a test. It can never catch a regression in what it claims to cover; the skip just hides the race.

## Fix

Replace the one-shot snapshot + `test.skip()` with web-first auto-waiting assertions:

```diff
- const initialCount = await allRows.count();
- if (initialCount < 1) { test.skip(); return; }
+ await expect(allRows.first()).toBeVisible();
+ const initialCount = await allRows.count();
+ expect(initialCount).toBeGreaterThan(0);
```
```diff
- if (!(await toggle.isVisible())) { test.skip(); return; }
+ await expect(toggle).toBeVisible();
```

The rows and the toggle always render (the toggle column is unconditional), so auto-waiting makes both tests run deterministically. Verified locally: the spec goes from 2 skips to 0 skips, all green. Happy to open a PR.

## Prevention (the part that keeps this from coming back)

The fix alone won't stop the next contributor, or a coding agent reaching for green, from doing the same. The testing philosophy should be documented in the template:

- A test must never `test.skip()` to dodge a timing race. Make it deterministic with auto-waiting (`expect(locator).toBeVisible()`, `expect.poll`), not a one-shot `count()`/`isVisible()` + skip.
- `test.skip()` is only for genuinely unsupported environments (a feature switched off, an external dependency absent), with a reason string that says which.
- Put this in a short TESTING/CONTRIBUTING note, and consider a lint rule that flags `test.skip()` adjacent to an `isVisible()`/`count()` read so it can't slip in unreviewed.

stickerdaniel:
Closes #508

Two tests in `e2e/admin-settings.spec.ts` called `test.skip()` when a non-retrying snapshot read came back empty: `filters recipients by type` skipped when a one-shot `allRows.count()` returned 0, and `toggles notification preference` skipped when `toggle.isVisible()` was false. Both reads can land in the window right after the `recipients-loading` flag clears but before the reactive table paints its rows, so on a slow run the tests silently skip and produce no signal instead of catching regressions.

The skip blocks are replaced with web-first auto-waiting assertions. The filter test now waits with `await expect(allRows.first()).toBeVisible()`, the unused `initialCount` snapshot is removed. The toggle test asserts `expect(email).toBeTruthy()` after extracting it from the row testid (the row is already auto-waited into view, so a missing testid is a real failure) and waits with `await expect(toggle).toBeVisible()` since the toggle column renders unconditionally. `AGENTS.md` gains a convention bullet in the E2E Playwright Tests section: never `test.skip()` to dodge a timing race, use auto-waiting assertions instead, and reserve `test.skip()` for genuinely unsupported environments with a reason string, pointing to the `hasGoogleOAuth` gate in `e2e/signin-last-used-badge.spec.ts` as the good example. The issue's optional lint rule flagging `test.skip()` next to one-shot `count()`/`isVisible()` reads is deferred as a follow-up since it needs heuristic AST matching.

`bun scripts/static-checks.ts e2e/admin-settings.spec.ts AGENTS.md` passes, and `bun run test:e2e -- e2e/admin-settings.spec.ts` passes with 0 skipped (8 passed, previously up to 2 conditional skips), stable across `--repeat-each=3` (22 passed).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -301,6 +301,7 @@ This project uses **PostHog** for product analytics with an optional **Cloudflar
 - Requires `.env.test` with `AUTH_E2E_TEST_SECRET` only. The Convex backend URL is auto-resolved from `.convex/.test-backend-url`; no `PUBLIC_CONVEX_URL` needed locally.
 - `AuthenticatedLayout` sets `data-hydrated` on `<html>` in `onMount`; `waitForAuthenticated` in `e2e/utils/auth.ts` waits for it so clicks happen after Svelte hydration, not on inert SSR markup.
 - See `.env.schema` for all available env vars with types and descriptions
+- **Never `test.skip()` to dodge a timing race.** Make tests deterministic with auto-waiting assertions (`expect(locator).toBeVisible()`, `expect.poll`) instead of one-shot `count()`/`isVisible()` snapshots followed by a skip. `test.skip()` is only for genuinely unsupported environments (a feature switched off, an external dependency absent), always with a reason string (good example: `e2e/signin-last-used-badge.spec.ts` gating on `hasGoogleOAuth`).
 - **CI E2E timing (CF Workers):** E2E tests only start after the CF Workers Build preview deployment completes (~2-3 min). The `e2e-preview-cf.yml` workflow triggers on the CF `check_run` event, extracts the preview URL, then runs Playwright against it. Results are posted as a commit status, not a check run. Total time from push to E2E result is ~7-8 min.
 
 #### Local e2e isolation
--- a/e2e/admin-settings.spec.ts
+++ b/e2e/admin-settings.spec.ts
@@ -113,15 +113,9 @@ test.describe('Admin Settings Page', () => {
 		// Wait for table to load
 		await expect.poll(async () => page.getByTestId('recipients-loading').count()).toBe(0);
 
-		// Get initial row count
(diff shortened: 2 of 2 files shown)


stickerdaniel:
* **#517** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/517?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/517?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #518, opened 2026-06-11
fix(support): Remove dead streaming state guards

### LINKED ISSUE #456: Support widget: hasPendingToolCalls guard is always false, streaming getters are dead
In src/lib/components/customer-support/support-thread-context.svelte.ts, `this.messages` is only ever assigned `[]` (lines 220 and 619). Consequences:

- `hasPendingToolCalls` (line 187) always returns false. It is the only tool-call guard on the chatbar send button (ai-chatbar.svelte:189), so the guard never engages.
- `streamingMessages`/`isStreaming` (line 172) are permanently empty/false. feedback-widget.svelte:280 consumes the dead getter.
- ChatCore.svelte.ts:108 carries the same vestigial messages-based getters, with filter drift between the two copies. Live streaming detection is ChatUIContext (ChatContext.svelte.ts:555), a third definition of the same concern.

proposed fix: derive these from the live display-message source the UI renders, or remove the dead getters and the disabled-clause; one canonical streaming-detection source.

stickerdaniel:
Closes #456

In `src/lib/components/customer-support/support-thread-context.svelte.ts` the `messages` field was only ever assigned `[]`, so the `hasPendingToolCalls`, `streamingMessages`, and `isStreaming` getters derived from it were constant false. The chatbar send button (`ai-chatbar.svelte`) carried a tool-call guard that could never engage, the widget send handler (`feedback-widget.svelte`) consumed the dead `isStreaming` getter, and `ChatCore.svelte.ts` duplicated the same vestigial messages-based getters with filter drift between the two copies, while live streaming detection already exists in `ChatUIContext.isStreaming`.

This removes the write-only `messages` fields and the dead getters from both `SupportThreadContext` and `ChatCore`, drops the constant `hasPendingToolCalls` clause from the chatbar send button, and tightens the `sendMessage` guard to `isSending || isAwaitingStream`. The widget send guard now uses `chatUIContext.isProcessing`, the same guard `ChatInput.svelte` already uses, which combines `isSending`, `isAwaitingStream`, and the live display-message based `ChatUIContext.isStreaming`, so streaming detection has one canonical source. `messages` is also removed from the `ExternalCoreAdapter` interface in `ChatRoot.svelte`, and the headless-core doc examples in `src/lib/chat/index.ts` and the context class doc no longer reference the removed members. All removed guard terms were constant false, so behavior is unchanged except the widget send guard now also catches the live streaming window.

`bun scripts/static-checks.ts` on the six changed files and `bun run test:unit` (486 tests) pass; a manual dev-server smoke test of the widget send flow was not run.

### DIFF
--- a/src/lib/chat/core/ChatCore.svelte.ts
+++ b/src/lib/chat/core/ChatCore.svelte.ts
@@ -213,7 +213,8 @@ export class ChatCore {
 				threadId = result.threadId;
 				threadCreated = result;
 
-				// Update threadId without clearing messages (for optimistic updates)
+				// Update threadId directly (setThread would clear streamCache, error,
+				// and the pagination cursors)
 				this.threadId = threadId;
 				this.isNewConversation = false;
 			}
--- a/src/lib/chat/core/ChatCore.svelte.ts
+++ b/src/lib/chat/core/ChatCore.svelte.ts
@@ -8,7 +8,6 @@
 import type { ConvexClient } from 'convex/browser';
 import type { FunctionReference } from 'convex/server';
 import type {
-	ChatMessage,
 	SendMessageOptions,
 	SendMessageResult,
 	ChatConfig,
@@ -66,9 +65,6 @@ export class ChatCore {
 	 *  Used as a {#key} to replay chip entrance animations. */
 	threadGeneration = $state(0);
 
-	// Messages state
-	messages = $state<ChatMessage[]>([]);
-
 	// Loading states
 	isLoading = $state(false);
 	isSending = $state(false);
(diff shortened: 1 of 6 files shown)


stickerdaniel:
* **#518** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/518?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/518?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #519, opened 2026-06-11
fix(scripts): Harden build script error handling

### LINKED ISSUE #499: Build scripts: unhandled rejections, uncaught JSON.parse on corrupt keys.json, silent cache-bypass loss in patch-cf-worker
Three robustness gaps in the build and deploy scripts, all low severity. Related to the build-script convention deviations in #473.

1. Bare async `main()` without `.catch()`. `scripts/deploy.ts:61` calls `main();` un-awaited and with no `.catch()`, so a step that throws instead of calling `process.exit(1)` relies on the runtime making unhandled rejections fatal. A reachable throw exists: `detectPlatform()` (called at `deploy.ts:30`) runs `sanitizeBranchAlias()`, which throws at `platform.ts:22`, bypassing the explicit `process.exit(1)` used elsewhere. The siblings `generate-convex.ts:274`, `model-eval.ts:98` and `static-checks.ts:368` all wrap `main().catch((err) => { console.error(err); process.exit(1); })`. Same gap in `scripts/template-setup.ts:319` (async `main()` declared at 187; a throw from `JSON.parse(read('package.json'))` at 261 or file I/O escapes) and `scripts/build-emails.ts:169` (async `buildEmails()` declared at 69; reject paths at `createViteServer` line 80, `ssrLoadModule` 84 and 87, `writeFileSync` 145, and `vite.close()` in the finally at 165 escape). On the currently pinned Bun an unhandled rejection does exit 1, so this is a consistency gap rather than a live bug.

2. `generate-convex.ts:216` runs `JSON.parse(fs.readFileSync(keysPath, 'utf-8'))` with no try/catch in the local-backend path. A truncated or corrupt `keys.json` (dev server killed mid-write, or a branch-switch race) throws an uncaught exception that bubbles to `main().catch` and exits 1, instead of degrading to offline validation the way the adjacent missing-keys branch (`208-214`) deliberately does. `keys.json` is written by the embedded backend, so a truncated file is a realistic state that branch was built to tolerate. Low impact since `bun run generate` is a standalone dev helper and runs nowhere in build, deploy, or CI.

3. `patch-cf-worker.ts:53-68` loses cache-bypass silently on cache-call drift. If adapter-cloudflare's worktop cache call changes shape to take extra args (for example `await cache(req, opts)`), `CACHE_LOOKUP_PATTERN` fails to match, the else-branch runs, and `applyMarkdownPatch` still returns a non-null patched string (static serving patched only), so the postbuild exits 0. The drift warning at line 55 is gated on the same `/await \w+\(req\)/` single-arg regex, so it also fails to fire for that exact drift, and the build succeeds with cache-bypass silently lost. Per the file's own docstring this breaks `Accept: text/markdown` content negotiation on non-prerendered pages.

proposed fix: wrap the four script entrypoints in `.catch((err) => { console.error(err); process.exit(1); })` to match the siblings, wrap the `keys.json` `JSON.parse` in a try/catch that warns and falls through to offline validation like the missing-keys branch, and in `patch-cf-worker` make the else-branch throw or return null when a cache-like call was detected so a build with lost cache-bypass fails loud instead of exiting 0.

stickerdaniel:
Closes #499

The build scripts had three robustness gaps. `scripts/deploy.ts`, `scripts/template-setup.ts` and `scripts/build-emails.ts` called their async entrypoints bare, so a reachable throw (for example `sanitizeBranchAlias()` via `detectPlatform()`) surfaced as an unhandled rejection instead of the explicit `process.exit(1)` used elsewhere. `scripts/generate-convex.ts` ran `JSON.parse` on `keys.json` without a guard, so a truncated file (dev server killed mid-write) crashed the script instead of degrading to offline validation like the adjacent missing-keys branch. And `scripts/patch-cf-worker.ts` lost the markdown cache-bypass silently when the worktop cache call drifted to a multi-arg shape, because both `CACHE_LOOKUP_PATTERN` and the single-arg drift-warning regex failed to match, so the build exited 0 with `Accept: text/markdown` negotiation broken on non-prerendered pages.

The three entrypoints are now wrapped in `.catch((err) => { console.error(err); process.exit(1); })`, matching `generate-convex.ts` and `model-eval.ts`. The `keys.json` read is hoisted into a try/catch that warns about a corrupt or truncated file and falls through to offline validation. In `applyMarkdownPatch`, the drift-detection regex is loosened from `await \w+\(req\)` to `await \w+\(req\b` so it catches multi-arg calls like `await r2(req, opts)` while still ignoring dotted calls like `server.respond(req, ...)`, and a detected drift now throws instead of warning and soft-falling back. The CLI entrypoint catches that error and exits 1 with the message. The silent static-serving fallback is preserved for workers without a cache layer, and a new test fixture with a drifted multi-arg cache call asserts the loud failure.

`bun scripts/static-checks.ts` on the six changed files passes, `bun run test:unit` passes (including the new drifted-cache fixture), and `bun run build:emails` and `bun run generate` both still exit 0.

### DIFF
--- a/scripts/build-emails.ts
+++ b/scripts/build-emails.ts
@@ -166,4 +166,7 @@ ${exportStatements}
 	}
 }
 
-buildEmails();
+buildEmails().catch((err) => {
+	console.error(err);
+	process.exit(1);
+});
--- a/scripts/deploy.ts
+++ b/scripts/deploy.ts
@@ -58,4 +58,7 @@ async function main(): Promise<void> {
 	console.log(`${colors.green}Deployment complete!${colors.reset}`);
 }
 
-main();
+main().catch((err) => {
+	console.error(err);
+	process.exit(1);
+});
--- a/scripts/generate-convex.ts
+++ b/scripts/generate-convex.ts
@@ -205,6 +205,7 @@ async function main() {
 			const stateId = computeLocalConvexStateId(PROJECT_DIR);
 			const keysPath = path.join(CONVEX_STATE_DIR, stateId, 'keys.json');
 
+			let keys: { adminKey: string } | null = null;
 			if (!fs.existsSync(keysPath)) {
 				console.warn(
 					`Warning: Local backend is running but no keys found at .convex/${stateId}/keys.json.`
@@ -213,7 +214,19 @@ async function main() {
 					'This can happen after switching branches. Falling back to offline validation.'
 				);
 			} else {
-				const keys = JSON.parse(fs.readFileSync(keysPath, 'utf-8'));
+				try {
+					keys = JSON.parse(fs.readFileSync(keysPath, 'utf-8'));
+				} catch {
+					console.warn(
+						`Warning: Could not parse .convex/${stateId}/keys.json (corrupt or truncated).`
+					);
+					console.warn(
+						'This can happen if the dev server was killed mid-write. Falling back to offline validation.'
+					);
+				}
+			}
+
+			if (keys) {
 				const { code, output } = runCommand('bunx', [
 					'convex',
 					'codegen',
--- a/scripts/patch-cf-worker.test.ts
+++ b/scripts/patch-cf-worker.test.ts
@@ -30,6 +30,18 @@ const WORKER_FIXTURE_NO_CACHE = `
     }
 `;
 
+// Fixture where the worktop cache call drifted to take extra args, so
+// CACHE_LOOKUP_PATTERN no longer matches but a cache lookup clearly exists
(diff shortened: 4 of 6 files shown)


stickerdaniel:
* **#519** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/519?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/519?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #520, opened 2026-06-11
docs(autumn): Note atmn push stays a manual step

### LINKED ISSUE #493: Autumn config: no guard the committed autumn.config.ts is live, and a third Pro allowance number in dead copy
Two config-hygiene gaps around `autumn.config.ts` that a fork inherits verbatim.

The Pro AI-chat allowance is stated as three different numbers. `autumn.config.ts:60` grants `aiChatMessages` `included: 500`, the plan JSDoc at `autumn.config.ts:46` says "200 AI chat messages/month", and the dead `ai_chat.pro_required.description` upsell copy (`en.json:1308`, `de.json:1307`, `es.json:1307`, `fr.json:1307`) tells users "Upgrade to Pro for 30 AI chat messages per month." Autumn enforces the config value, so the in-product upsell understates the real entitlement by a wide margin. #455 already tracks the 500-vs-200 docstring and notes `ai_chat.pro_required` is referenced nowhere in code. The new fact here is that the key's "30" is a third value contradicting both, so removing the key as #455 proposes also removes the contradiction.

Separately there is no guard that the committed config was actually pushed. `AGENTS.md` requires `bunx atmn push` after any `autumn.config.ts` edit because the config stays inert until pushed, but `static-checks.ts:343` runs only `bun atmn preview`, a local render that never diffs against or pushes to the sandbox or prod deployment. No workflow references `atmn`. A fork can edit limits, commit, and silently run against stale live entitlements with nothing flagging the divergence. The Autumn CLI exposes no read-only diff primitive for building a true drift guard, and auto-pushing from CI would be its own anti-pattern, so this stays a manual step that is easy to forget.

proposed fix: remove the dead `ai_chat.pro_required` block in all four locale files alongside the #455 cleanup, and add a short note near the `atmn preview` call in `static-checks.ts` (and in the `AGENTS.md` Autumn section) that `atmn push` is a required manual step `preview` does not cover.

stickerdaniel:
Closes #493

`AGENTS.md` requires `bunx atmn push` after any `autumn.config.ts` edit because the config stays inert until pushed, but the only automated touchpoint is `bun atmn preview` in `scripts/static-checks.ts`, a local render that never diffs against or pushes to the live deployment. Nothing in the docs or the script says so, which makes it easy for a fork to edit limits, commit, pass all checks, and silently run against stale live entitlements.

This expands the comment above the `atmn preview` call in `scripts/static-checks.ts` to state that it only renders the local config and that `bunx atmn push` (sandbox) and `bunx atmn push -p` (prod) remain a required manual step no automated check covers, and adds the same caveat to the `AGENTS.md` Autumn Billing Config section. Per the issue, no auto-push or drift guard is added: the only diff the Autumn CLI ships is the hidden debug-only `test-diff` command, which requires auth, always exits 0, and prints human-readable output, so there is no stable primitive to build a guard from, and auto-pushing from CI would be its own anti-pattern. The third Pro allowance number in the dead `ai_chat.pro_required` locale copy and the 200-message docstring were already removed in #506, so no locale or config changes are needed here.

`bun scripts/static-checks.ts scripts/static-checks.ts AGENTS.md` passes, including the unchanged `atmn preview` step.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -162,7 +162,7 @@ After modifying `autumn.config.ts`, ALWAYS push changes to Autumn:
 - `bunx atmn push` — Push config to sandbox
 - `bunx atmn push -p` — Push config to production
 
-Without pushing, the config change only exists locally and has no effect.
+Without pushing, the config change only exists locally and has no effect. Pre-commit/static-checks only runs `atmn preview` (a local render that never diffs against or pushes to the live deployment), so no automated check covers pushing; it stays a manual step.
 
 ### Tolgee CLI
 
--- a/scripts/static-checks.ts
+++ b/scripts/static-checks.ts
@@ -342,8 +342,10 @@ async function main(): Promise<void> {
 		// `atmn preview` only renders plans from the local autumn.config.ts; it never
 		// diffs against or pushes to the live deployment. After any config edit,
 		// `bunx atmn push` (sandbox) / `bunx atmn push -p` (prod) is a required manual
-		// step that no automated check covers (the CLI has no read-only diff primitive
-		// to build a drift guard from, and auto-pushing from CI would be worse).
+		// step that no automated check covers (the CLI's only diff is the hidden
+		// debug-only `test-diff`, which requires auth, always exits 0, and prints
+		// human-readable output, so there is no stable primitive to build a drift
+		// guard from; auto-pushing from CI would be worse).
 		printHeader(step, 'Autumn config');
 		runCommand('bun', ['atmn', 'preview']);
 		console.log('\n');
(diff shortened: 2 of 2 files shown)


stickerdaniel:
* **#520** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/520?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/520?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #521, opened 2026-06-11
fix(auth): Revoke sessions on password reset

### LINKED ISSUE #453: Password config: reset keeps other sessions alive, no max length, complexity client-only
src/lib/convex/auth.ts:323 sets only `minPasswordLength: 10`. Three gaps in the same config block:

- `revokeSessionsOnPasswordReset` defaults to false, so resetting a password after an account takeover leaves the attacker's session alive. The settings change-password flow already opts into `revokeOtherSessions: true`; the reset flow does not.
- no `maxPasswordLength`, so hashing cost is unbounded.
- uppercase/lowercase/digit complexity lives only in the client Valibot schema (src/lib/schemas/password.ts, copied into signin/schema.ts, reset-password/schema.ts and settings/password-schema.ts), so any direct API call skips it.

proposed fix:

```diff
 			minPasswordLength: 10,
+			maxPasswordLength: 256,
+			revokeSessionsOnPasswordReset: true,
```

and decide whether complexity is policy (enforce server-side) or a UX hint (document it as advisory).

stickerdaniel:
Closes #453

The `emailAndPassword` block in `src/lib/convex/auth.ts` only set `minPasswordLength: 10`. Better Auth's `revokeSessionsOnPasswordReset` defaults to false, so completing a password reset after an account takeover left the attacker's existing sessions alive, while the settings change-password flow already opts into `revokeOtherSessions: true`. Nothing pinned the reset behavior in tests either.

This sets `revokeSessionsOnPasswordReset: true` so a completed reset deletes all of the user's sessions (verified against better-auth 1.6.9: the reset route calls `internalAdapter.deleteSessions`, which the Convex adapter backs with `deleteMany`). It also sets `maxPasswordLength: 256`. Better Auth already caps passwords at 128 by default, so this raises the cap for long passphrases and aligns with the `email-and-password-best-practices` skill example rather than fixing unbounded hashing as the issue suggested. A new E2E test in `e2e/forgot-reset-password.spec.ts` pins the revocation: it signs up a dedicated user, establishes a second session in a separate cookie jar, completes the reset through the UI, and asserts `/api/auth/get-session` returns null for the other session afterwards. Since reset emails are skipped for `@e2e.example.com` addresses, a new secret-gated mutation `getPasswordResetToken` in `src/lib/convex/tests.ts` reads the token Better Auth stored in the verification model, dead code in production like the other `AUTH_E2E_TEST_SECRET` helpers. Whether the uppercase/lowercase/digit complexity rules should be enforced server-side or documented as advisory is left open, as the issue explicitly defers that decision.

`bun scripts/static-checks.ts` on the changed files, `bun run check:convex`, `bun run test:unit` (486 tests), and `bun run test:e2e -- e2e/forgot-reset-password.spec.ts` (9 tests, including the new revocation test) all pass locally.

### DIFF
--- a/e2e/forgot-reset-password.spec.ts
+++ b/e2e/forgot-reset-password.spec.ts
@@ -1,7 +1,23 @@
 import { test, expect } from '@playwright/test';
+import 'varlock/auto-load';
+import { ConvexHttpClient } from 'convex/browser';
+import { api } from '../src/lib/convex/_generated/api';
+import { resolveConvexUrl } from './utils/convex-url';
+import { resolveSiteUrl } from './utils/site-url';
+import { getPreviewBypass } from './utils/preview-bypass';
 
 // This test runs without auth state - tests unauthenticated behavior
 test.use({ storageState: { cookies: [], origins: [] } });
+
+const testSecret = process.env.AUTH_E2E_TEST_SECRET!;
+
+const getConvexClient = () => {
+	const convexUrl = resolveConvexUrl();
+	if (!convexUrl) {
+		throw new Error('Convex URL not configured (set PUBLIC_CONVEX_URL or start local dev server)');
+	}
+	return new ConvexHttpClient(convexUrl);
+};
 
 test.describe('Forgot Password', () => {
 	test('shows validation error for invalid email', async ({ page }) => {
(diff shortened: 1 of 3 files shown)


stickerdaniel:
* **#521** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/521?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/521?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #522, opened 2026-06-11
refactor(auth): Unify viewer query and JWT decoder

### LINKED ISSUE #471: Duplicate auth helpers: getCurrentUser vs users.viewer, two JWT decoders
Two pairs of duplicated auth plumbing:

- `auth.getCurrentUser` (auth.ts:383) and `users.viewer` (users.ts:12) are identical wrappers over `authComponent.getAuthUser(ctx)`. Consumers split arbitrarily: +layout.server.ts:91 and command-menu.svelte:48 use one, ai-chat/community-chat/admin pages use the other.
- hooks.server.ts:43 and +layout.server.ts:19 both decode the same JWT cookie with divergent shapes and typing; the admin role gate relies on the looser of the two.

proposed fix: keep `api.users.viewer` (adding its missing returns validator), delete `getCurrentUser`, update the two call sites; extract one typed payload decoder into src/lib/server/ and use it from both files.

stickerdaniel:
Closes #471

`auth.getCurrentUser` and `users.viewer` were identical wrappers over `authComponent.getAuthUser(ctx)`, with consumers split arbitrarily between the two. In parallel, `hooks.server.ts` and `+layout.server.ts` each decoded the same JWT cookie with their own helper and divergent payload shapes, and the admin role gate relied on the looser of the two decoders. `users.viewer` also had no returns validator and a JSDoc claiming it returns null when unauthenticated, even though `getAuthUser` throws `ConvexError('Unauthenticated')`.

`api.users.viewer` is now the single viewer query. It gains a returns validator built from the Better Auth local-install schema (`v.object({ _id: v.string(), _creationTime: v.number(), ...tables.user.validator.fields })`) so the validator stays in sync with the schema, and its JSDoc documents the throw, which the JWT payload fallback in `+layout.server.ts` relies on. The two `api.auth.getCurrentUser` call sites in `+layout.server.ts` and `command-menu.svelte` are migrated and the duplicate query is deleted. JWT decoding moves to a single typed `decodeJwtPayload` in `src/lib/server/jwt.ts`, used by both server files and covered by a co-located unit test. It returns null when the `sub` claim is missing, matching the stricter of the two previous decoders, so the admin role gate gets tighter rather than looser.

`bun scripts/static-checks.ts` on the changed files, `bun run check:convex`, and `bun run test:unit` pass. `bun run test:e2e` against the isolated local stack passes 57 of 67 specs; the 8 failures (axe navigation races and AI chat load timeouts) are unrelated parallel-run flakes and all pass on a targeted re-run.


stickerdaniel:
* **#522** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/522?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/522?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #523, opened 2026-06-11
test(i18n): Sync wrangler.toml run_worker_first with languages

### LINKED ISSUE #472: wrangler.toml run_worker_first is a fourth untested copy of the language list
The supported-language list now lives in four places: src/lib/i18n/languages.ts (canonical), svelte.config.js, the Tolgee config, and wrangler.toml:13 (`run_worker_first = ["/en", "/en/*", ...]`). prerender-sync.test.ts asserts the first three stay in sync and does not parse wrangler.toml. Adding a language would silently skip worker-first routing for the new locale, breaking `Accept: text/markdown` negotiation on its marketing pages.

proposed fix: extend prerender-sync.test.ts to parse the run_worker_first entries, derive the unique language codes, and assert equality with the canonical list; add a pointer comment in wrangler.toml like the one in svelte.config.js.

stickerdaniel:
Closes #472

The supported-language list lives in four places: `src/lib/i18n/languages.ts` (canonical), `svelte.config.js`, the Tolgee imports in the root layout, and the `run_worker_first` array in `wrangler.toml`. `scripts/prerender-sync.test.ts` only asserted the first three stay in sync, so adding a language would silently skip worker-first routing for the new locale and break `Accept: text/markdown` content negotiation on its prerendered marketing pages.

The sync test now reads `wrangler.toml`, regex-extracts the `run_worker_first` array, and compares the entries against the set derived from `SUPPORTED_LANGUAGES` (both the bare `/<lang>` prefix and the `/<lang>/*` wildcard per language, so a missing wildcard fails too, slightly stricter than the unique-codes check the issue proposed). The test asserts the regex matched at all, so restructuring `wrangler.toml` fails loudly instead of passing vacuously. The comment above `run_worker_first` now points at `SUPPORTED_LANGUAGES` in `src/lib/i18n/languages.ts` and at `scripts/prerender-sync.test.ts`, mirroring the pointer comment in `svelte.config.js`.

`bun run test:unit` passes (51 files, 487 tests), `bun scripts/static-checks.ts scripts/prerender-sync.test.ts` passes, and a negative check (removing `"/fr/*"` from `wrangler.toml`) makes the new test fail as expected.

### DIFF
--- a/scripts/prerender-sync.test.ts
+++ b/scripts/prerender-sync.test.ts
@@ -1,3 +1,5 @@
+import * as fs from 'node:fs';
+import * as path from 'node:path';
 import { describe, expect, it } from 'vitest';
 import { SUPPORTED_LANGUAGES } from '../src/lib/i18n/languages';
 import { SUPPORTED_LOCALES } from '../src/lib/convex/i18n/translations';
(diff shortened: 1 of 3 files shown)


stickerdaniel:
* **#523** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/523?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/523?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #525, opened 2026-06-11
fix(seo): Bake site origin into prerendered tags

### LINKED ISSUE #479: SEOHead bakes http://sveltekit-prerender into canonical, hreflang, og:url and og:image on prerendered pages
`SEOHead.svelte:32` derives every absolute SEO URL from `origin = $derived(page.url.origin)`. On prerendered marketing routes that origin is SvelteKit's build-time placeholder `http://sveltekit-prerender`, so the shipped static HTML for every prerendered page carries `http://sveltekit-prerender` in `rel=canonical` (line 65), `og:url` (66), `og:image` and `twitter:image` (sourced from `absoluteImageUrl` at line 48, emitted at 71 and 77), all four hreflang alternates (85), and `x-default` (93).

Crawlers read the static HTML before hydration, so in production Google canonicalization and hreflang targeting point at a nonexistent host and social link previews fetch a broken image URL. The build succeeds, so a fork inherits this silently.

The build output confirms it: `.svelte-kit/output/prerendered/pages/{en,de,es,fr}.html` contains `http://sveltekit-prerender` in those tags. `svelte.config.js` sets no `kit.prerender.origin` (the `prerender` block at lines 36-38 has only `entries` and `handleMissingId`), so the SvelteKit default placeholder applies. `PUBLIC_SITE_URL` already exists at `.env.schema:122` but `SEOHead` never imports it. The `sitemap.xml` and `robots.txt` `+server.ts` endpoints run at request time and use the real `url.origin`, so this is isolated to the prerendered HTML SEO tags.

Same prerender-freeze failure class as #452 (frozen `page.data.viewer` on the marketing home). #474 also edits this file for the `LEGAL_CONFIG.brandName` suffix.

proposed fix: build absolute URLs from `PUBLIC_SITE_URL` (`$env/static/public`) and fall back to `page.url.origin` at runtime. `PUBLIC_SITE_URL` is `@optional` and defaults to localhost under test, so prefer the env value when it is set and keep the runtime origin otherwise:

```js
import { PUBLIC_SITE_URL } from '$env/static/public';
let origin = $derived(PUBLIC_SITE_URL || page.url.origin);
```

stickerdaniel:
Closes #479

`SEOHead.svelte` derived every absolute SEO URL from `page.url.origin`, which is SvelteKit's build-time placeholder `http://sveltekit-prerender` on prerendered routes. The shipped static HTML for every prerendered marketing page therefore carried that nonexistent host in `rel=canonical`, `og:url`, `og:image`, `twitter:image` and all hreflang alternates. Crawlers read the static HTML before hydration, so Google canonicalization and hreflang targeting pointed at a dead host and social previews fetched a broken image URL, silently in any fork since the build succeeds.

`SEOHead` now imports `PUBLIC_SITE_URL` from `$env/static/public`, strips a trailing slash, and prefers it over `page.url.origin`, which stays as the runtime fallback when the var is unset. `computeBuildEnv` in `scripts/deploy/steps.ts` derives `PUBLIC_SITE_URL` from `platform.siteUrl` when it is not set explicitly, so CF Workers and Vercel builds bake the real deployment origin into the prerendered HTML while an explicitly configured value (for example a custom domain) always wins. On Vercel, `detectPlatform` prefers `VERCEL_PROJECT_PRODUCTION_URL` for production builds, since `VERCEL_URL` is the per-deployment generated host that changes on every deploy; previews keep the deployment URL. The `PUBLIC_SITE_URL` comment in `.env.schema` now documents this role, with `src/env.d.ts` regenerated via `varlock typegen`, and new tests in `scripts/deploy/steps.test.ts` and `scripts/deploy/platform.test.ts` cover the derive, no-overwrite, null `siteUrl` and Vercel production-domain cases.

`bun scripts/static-checks.ts` on the changed files, `bun run test:unit` (494 tests), and a build proof both ways pass: `PUBLIC_SITE_URL=https://example.com bun run build` produces prerendered pages with zero `sveltekit-prerender` occurrences and `https://example.com` in canonical, hreflang and `og:image`, while a plain `bun run build` without the var still succeeds with the runtime-origin fallback.

### DIFF
--- a/.env.schema
+++ b/.env.schema
@@ -117,7 +117,10 @@ PUBLIC_SNAPDOM_PROXY_URL=
 # @sensitive @optional
 AUTH_E2E_TEST_SECRET=
 
-# Site URL for test requests (defaults to localhost:5173)
+# Canonical site origin (e.g. https://example.com). Baked into prerendered SEO
+# tags (canonical, hreflang, og:url, og:image) at build time; also the E2E test
+# base URL in CI (defaults to localhost:5173). The deploy script derives it from
+# the platform when unset; set it explicitly for custom domains.
 # @type=url @optional
 PUBLIC_SITE_URL=
 
--- a/scripts/deploy/platform.test.ts
+++ b/scripts/deploy/platform.test.ts
@@ -1,5 +1,5 @@
-import { describe, expect, it } from 'vitest';
-import { sanitizeBranchAlias } from './platform';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { detectPlatform, sanitizeBranchAlias } from './platform';
 
 describe('sanitizeBranchAlias', () => {
 	it('uses the full alias budget for short worker names', () => {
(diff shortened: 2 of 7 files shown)


stickerdaniel:
* **#525** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/525?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/525?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #526, opened 2026-06-11
refactor(hooks): Scope shared $state per request

### LINKED ISSUE #500: `$effect` syncs props into state and module-scope `$state` singletons risk SSR cross-request leaks
Two Svelte 5 patterns in the template are worth tightening.

First, `$effect` copies a reactive value into editable form `$state`. The effect re-runs on any reactive change to its source and silently overwrites the user's unsaved typing, and `$derived` is the idiomatic way to mirror a value.

- `account-settings.svelte:33-36` copies the `user` prop into the `name` and `image` `$state` fields. `user` is reactive (the parent passes `$derived(data.user)` at `settings/+page.svelte:22`), so a new `data.user` reference (for example a post-save load invalidation) clobbers in-progress input or the avatar preview. `data.user` is SSR load data and not a live subscription, so the data-loss window is narrow.
- `signin/+page.svelte:90-94` copies the shared reactive `authFlow.email` into `signInData.email` (bound at `SignInForm.svelte:132`); a change to `authFlow.email` clobbers the typed email.
- `signup/+page.svelte:93-97` copies `authFlow.email` into `signUpData.email` (bound at `SignUpForm.svelte:121`); same shape.
- `forgot-password/+page.svelte:64-68` copies `authFlow.email` into `formData.email` (bound at `forgot-password/+page.svelte:200`); the `&& !formData.email` guard already prevents the overwrite.

Second, mutable module-level `$state` singletons. `auth-flow.svelte.ts:5-9` exports `authFlow`, a module-scope class whose `email` field is `$state('')`. The SvelteKit "Avoid shared state on the server" doc warns that state in a shared module leaks between users during SSR. Nothing leaks today because every read and write lives in client-only `$effect` blocks and the SSR inputs bind to local empty-string `$state`, but auth routes are SSR'd, so as a reference template any fork that prefills an SSR-rendered email input from `authFlow.email` would leak one user's entered email into another request. The same pattern appears in two more singletons:

- `admin-support-ui.svelte.ts:1-13` exports `adminSupportUI` whose `detailsOpen = $state(false)` is read during SSR via `bind:open` on `Sheet.Root` and `Drawer.Root` (`admin/support/+page.svelte:281` and `:290`), so concurrent admin requests share one value.
- `use-haptic.svelte.ts:101` exports `haptic` with `$state` fields, safe today because every write is browser-guarded and the state never renders into SSR HTML.

proposed fix: replace the prop-to-state effects with `$derived` (or initialize the field once and let the input own it afterward), and move the shared singletons into Svelte context via `runed`'s `Context` so each SSR request gets its own instance.

stickerdaniel:
Closes #500

Module-scope `$state` singletons (`authFlow` in `auth-flow.svelte.ts`, `adminSupportUI` in `admin-support-ui.svelte.ts`) live in shared module state, which the SvelteKit docs warn leaks between users during SSR. `adminSupportUI.detailsOpen` is read during SSR via `bind:open` on the support details Sheet and Drawer, so concurrent admin requests shared one value, and any fork that prefills an SSR-rendered input from `authFlow.email` would leak one user's email into another request. Separately, several components copied reactive values into editable form `$state` inside `$effect`, which re-runs on any reactive change to its source and silently overwrites unsaved typing.

Both singletons are now provided through runed `Context`. The `(auth)` layout sets a per-request `AuthFlowManager` instance. Since the layout survives client-side navigation between the auth pages, the email is still remembered when switching between sign-in, sign-up, and forgot-password. One behavior change: navigating out of the auth group and back now resets the remembered email, where the old module singleton kept it for the whole SPA session. This is an accepted, privacy-positive trade-off, since the email no longer lingers in memory after leaving the auth flow. The admin support page sets an `AdminSupportUIManager` instance that `thread-chat.svelte` reads from context, with all call sites unchanged. The inbound `$effect` blocks that copied `authFlow.email` into form state on `signin`, `signup`, and `forgot-password` are deleted; each form initializes its email field once from `authFlow.email` and owns it afterward, while the outbound sync effect keeps recording typed emails. In `account-settings.svelte` the prop-to-state `$effect` is replaced with writable `$derived` for `name` and `image`, so the form re-syncs when `user` changes and `bind:value`, the upload preview, and the remove handler keep working. The `haptic` singleton stays module-scoped with an intentional anti-pattern comment: every write of caller-controlled data is browser-guarded, the remaining writes set constants with no per-request data, and nothing is rendered into SSR HTML, so migrating its many call sites to context would add churn without removing any leak.

`bun scripts/static-checks.ts` on all changed files, `bun run test:unit` (486 tests), and the auth and support Playwright specs (`forgot-reset-password`, `invalid-auth`, `signin-invalid-credentials`, `signin-last-used-badge`, `signout`, `support-migration`) via `bun run test:e2e` all pass.

### DIFF
--- a/src/lib/hooks/admin-support-ui.svelte.ts
+++ b/src/lib/hooks/admin-support-ui.svelte.ts
@@ -1,4 +1,13 @@
-class AdminSupportUIManager {
+import { Context } from 'runed';
+
+/**
+ * UI state for the admin support details overlay (Sheet/Drawer).
+ *
+ * Provided via context from the admin support page so each SSR request gets
+ * its own instance; `detailsOpen` is read during SSR via `bind:open`, so a
+ * module-scope singleton would be shared across concurrent requests.
+ */
+export class AdminSupportUIManager {
 	detailsOpen = $state(false);
 
 	toggle() {
@@ -10,4 +19,4 @@ class AdminSupportUIManager {
 	}
 }
 
-export const adminSupportUI = new AdminSupportUIManager();
+export const adminSupportUIContext = new Context<AdminSupportUIManager>('admin-support-ui');
--- a/src/lib/hooks/auth-flow.svelte.ts
+++ b/src/lib/hooks/auth-flow.svelte.ts
@@ -1,9 +1,14 @@
+import { Context } from 'runed';
+
 /**
  * Shared email state across auth pages (sign-in, sign-up, forgot-password)
- * This improves UX by remembering the email when switching between auth flows
+ * This improves UX by remembering the email when switching between auth flows.
+ *
+ * Provided via context from the (auth) layout so each SSR request gets its own
+ * instance instead of sharing a module-scope singleton across requests.
  */
-class AuthFlowManager {
+export class AuthFlowManager {
 	email = $state('');
 }
 
-export const authFlow = new AuthFlowManager();
+export const authFlowContext = new Context<AuthFlowManager>('auth-flow');
--- a/src/lib/hooks/use-haptic.svelte.ts
+++ b/src/lib/hooks/use-haptic.svelte.ts
@@ -97,4 +97,8 @@ export class UseHaptic {
 	}
 }
 
+// Intentional module-scope singleton: SSR-safe because every $state write is
+// browser-guarded and no field is rendered into SSR HTML, so nothing can leak
+// across requests. Kept as a singleton (instead of context) for ergonomic
+// imports across its many call sites.
 export const haptic = new UseHaptic();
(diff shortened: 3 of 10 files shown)


stickerdaniel:
* **#526** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/526?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/526?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #527, opened 2026-06-11
fix(a11y): Honor prefers-reduced-motion in transitions

### LINKED ISSUE #475: Svelte fly/slide transitions ignore prefers-reduced-motion
Svelte transition directives run on the JS clock, so `prefers-reduced-motion` CSS rules cannot disable them. Position-moving transitions without a reduced-motion guard:

- [ ] ui/field/field-error.svelte:29: shared primitive, animates in nearly every form
- [ ] ai-elements/conversation/ConversationScrollButton.svelte:42
- [ ] ui/sliding-header/sliding-header.svelte:66, 102: already has a `skipTransition` escape hatch, wired only to deep-link state
- [ ] (auth)/reset-password/+page.svelte:192, 201 and (auth)/forgot-password/+page.svelte:173, 182: slide banners
- [ ] weaker (scale only): ui/copy-button/copy-button.svelte:50, ui/chat/chat-list.svelte:35

proposed fix: gate offset/duration on a reduced-motion check (`useReducedMotion()` from motion-sv, already a dependency, or a runed MediaQuery) and keep the opacity fade, e.g. `fly={{ y: reduce ? 0 : 8, duration: reduce ? 0 : 150 }}`; OR the sliding-header check into its existing skipTransition.

stickerdaniel:
Closes #475

Svelte transition directives animate on the JS clock, so the `prefers-reduced-motion` CSS media query cannot disable them. Every `fly` and `slide` in the app moved elements for users who asked the OS to reduce motion: the shared `Field.Error` primitive that animates in nearly every form, the forgot/reset-password banners, the AI chat scroll-to-bottom button, the support widget header view swap, and the `copy-button` scale pop.

The fix gates the movement parameters on the built-in `prefersReducedMotion` media query from `svelte/motion` and keeps the opacity fade. Fly offsets (`x`, `y`) and the scale `start` become neutral under reduced motion, so state changes still read as a crossfade without translation or zoom, and the `slide` banners and field errors get `duration: 0` so they appear instantly since slide is pure height movement. In `sliding-header.svelte` the media query is combined with the existing `skipTransition` deep-link escape hatch via a `noMotion` derived: both zero the offsets, while only `skipTransition` also zeroes the duration so deep links stay instant. Transition parameters are evaluated lazily when the transition plays, so the inline `prefersReducedMotion.current` reads stay reactive. The `ui/chat/chat-list.svelte` item from the issue is obsolete, that component no longer exists.

`bun scripts/static-checks.ts` on the six changed files and `bun run test:unit` (486 tests) pass. Not verified by hand in a browser with reduced-motion emulation since the change only zeroes offsets and durations on existing transitions.


stickerdaniel:
* **#527** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/527?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/527?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #528, opened 2026-06-11
fix(ui): Add error branches to useQuery consumers

### LINKED ISSUE #488: Admin/support useQuery consumers have no error branch; a query throw shows infinite skeletons
convex-svelte's `useQuery` exposes both a `data` and an `error` field, but no consumer in `src/routes` ever reads `.error`. Every one derives loading from `!data`, so a transient Convex or network throw leaves `data` undefined forever and the UI stays in its loading state with no message and no recovery path.

The admin dashboard is the clearest case: `admin/dashboard/+page.svelte:23-26` does `useQuery(api.admin.queries.getDashboardMetrics, {})` then `isLoading = $derived(!metrics.data)`. On a query throw `metrics.data` stays undefined and the dashboard shows skeleton cards indefinitely. The page has no `+page.server.ts`, no `<svelte:boundary>`, and no `+error.svelte`, so nothing recovers it. `AGENTS.md:459-466` prescribes this loading-only pattern using `getDashboardMetrics` as its literal example, so forks inherit the trap.

This propagates from the shared cursor-table factory at `create-convex-cursor-table.svelte.ts:382-402`: `listQuery`/`countQuery` come from `useQuery`, `isLoading` derives from `listQuery.isLoading`, and `listQuery.error`/`countQuery.error` are never read. The returned state object (`663-714`) exposes no error field at all, so every consumer table is stuck at `isLoading=true` on a throw. Consumers that inherit this:

- `admin/users/+page.svelte:190, 560-605` renders skeleton rows whenever `usersTable.isLoading`; a transient `getUsers` failure shows infinite skeletons.
- `admin/settings/notification-recipients-table.svelte:112, 308-345` has the same shape; a query error keeps `isLoading` true forever.

Standalone `useQuery` consumers with the same gap:

- `admin/support/+page.svelte:66-78` drives `isLoading` off `threadsQuery` (usePaginatedQuery) without reading `threadsQuery.error`; a first-page error leaves the inbox showing skeletons.
- `admin/support/thread-chat.svelte:96-102, 124` opens without an `initialThread`, so `isLoading = !initialThread && !thread` and `getThreadForAdmin`'s error is unread, rendering the skeleton header indefinitely.
- `admin/support/thread-details.svelte:36-41, 136` renders via `{#if thread}` with no `{:else}`; a query error leaves the panel blank.
- `admin/settings/founder-welcome-card.svelte:19, 127-128` uses `{#if !result.data}` for an invisible placeholder and never reads `result.error`, so a query error sticks on the placeholder.
- `customer-support/threads-overview.svelte:31-42, 219, 286` derives `isLoading` from `threadsQuery.isLoading` and ignores `error`, leaving the support widget's thread area permanently blank.

This is the client-side `useQuery` reactive-error seam, separate from #458, which covers server-side SSR loads missing try-catch.

proposed fix: thread an `error` field through the cursor-table factory (read `listQuery.error`/`countQuery.error`, expose `get error()`) and have `ConvexCursorTableShell` render a retry-able error row when set, then read `.error` in the standalone consumers above to swap the skeleton for an inline error state. Update the `AGENTS.md` pattern-3 example to show the error branch alongside the loading one so forks copy it.

stickerdaniel:
Closes #488

convex-svelte's `useQuery` and `usePaginatedQuery` expose an `error` field, but no consumer in `src/routes` read it. Every consumer derived its loading state from `!data` or from `isLoading` alone, so a transient Convex or network throw left `data` undefined forever and the UI stayed on skeletons (admin dashboard, users table, notification recipients table, support inbox, thread chat header) or went silently blank (thread details, founder welcome card, customer support widget) with no message and no recovery path.

The cursor-table factory `create-convex-cursor-table.svelte.ts` now derives an `error` from `listQuery.error ?? countQuery.error` and exposes it on the returned state. Both table consumers render a localized full-width error row (`admin-users-error`, `recipients-error`) ahead of the empty state, which previously swallowed a first-load failure because `isLoading` turns false on a throw while `rows` is still empty. After a successful load `rows` can stay non-empty from the page cache, so on a later subscription error the error row intentionally takes priority over the stale rows. The standalone consumers read `.error` directly: the admin dashboard replaces the metric cards with an inline error and derives loading from `metrics.isLoading` instead of `!metrics.data`, the support inbox passes the paginated query error into the thread list (only when nothing is loaded yet, since load-more failures already have the `InfiniteLoader` retry snippet), the thread chat header and the details panel swap their skeleton or blank state for an error message, the founder welcome card footer shows the error instead of its invisible placeholder button, and the customer support widget shows it instead of the greeting. No retry plumbing is added because the Convex subscription stays live after an error, so every error state self-heals once the query recovers. The copy is localized in all four locales via `common.load_error` and `support.thread.load_error`, and the `AGENTS.md` pattern-3 example now includes the error branch so forks copy it.

`bun scripts/static-checks.ts` on all changed files passes, `bun run test:unit` passes (486 tests), and `bun run test:e2e e2e/admin-users-table.spec.ts e2e/admin-settings-table.spec.ts e2e/admin-settings.spec.ts` passes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -440,11 +440,21 @@ For each data need in a new route, pick the right pattern:
 3. **Real-time data that can show a loading state?**
    → Skip server fetch, use `useQuery` directly (no `initialData`):
 
-   ```ts
-   const metrics = useQuery(api.admin.queries.getDashboardMetrics, {});
+   ```svelte
+   <script lang="ts">
+   	const metrics = useQuery(api.admin.queries.getDashboardMetrics, {});
+   	const isLoading = $derived(metrics.isLoading);
+   </script>
+
+   {#if metrics.error}
+   	<p class="text-destructive"><T keyName="common.load_error" /></p>
+   {:else}
+   	<!-- skeleton while isLoading, data when resolved -->
+   {/if}
    ```
 
    Shows skeleton/loading state until data arrives. Use for admin panels, secondary data.
+   Always render an error branch from `metrics.error` (localized via Tolgee) and derive loading from `isLoading`, never from `!metrics.data`: on a query throw `data` stays `undefined` forever, so a `!data` loading state shows infinite skeletons. No retry plumbing needed, the Convex subscription stays live and the error state self-heals when the query recovers.
 
 4. **Billing/subscription checks?**
    → `useCustomer()` from `@stickerdaniel/convex-autumn-svelte/sveltekit`
--- a/src/i18n/de.json
+++ b/src/i18n/de.json
@@ -636,6 +636,7 @@
 		"confirm": "Bestätigen",
 		"confirm_input_placeholder": "Geben Sie \"{text}\" ein, um zu bestätigen.",
 		"delete": "Löschen",
+		"load_error": "Daten konnten nicht geladen werden. Bitte versuchen Sie es später erneut.",
 		"save": "Speichern",
 		"you": "Sie"
 	},
@@ -1086,6 +1087,7 @@
 			"question_text": "Warum SaaS Starter?"
 		},
 		"thread": {
+			"load_error": "Unterhaltungen konnten nicht geladen werden. Bitte versuchen Sie es später erneut.",
 			"new_conversation": "Neue Unterhaltung"
 		},
 		"time": {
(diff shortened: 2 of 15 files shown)


stickerdaniel:
* **#528** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/528?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/528?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #534, opened 2026-06-11
refactor(loading-bar): Use runed AnimationFrames

### LINKED ISSUE #390: loading-bar.svelte: requestAnimationFrame loop + state mutations in $effect
## Problem

\`src/lib/components/ui/loading-bar/loading-bar.svelte\` has two related drift signals:

### 1. Manual \`requestAnimationFrame\` ID tracking
Lines 38, 72, 83, 107, 114 own a \`rafId\` variable, guard against double-calls, and manually \`cancelAnimationFrame\` in three places. Runed's \`AnimationFrames\` utility handles this loop with FPS-limiting + auto cleanup:

\`\`\`ts
import { AnimationFrames } from 'runed';
const frames = new AnimationFrames(renderLoop);
$effect(() => {
  if (isLoading) frames.start();
  else frames.stop();
});
\`\`\`

### 2. \`$state\` mutations inside \`$effect\`
Lines 43-92: multiple \`$effect\` blocks mutate \`progressStyle\` / \`isLoading\` / \`progress\` directly. The Svelte MCP autofixer flags every one with *"stateful variable assigned inside an \$effect which is generally considered a malpractice."*

This is fixable by either splitting the effects so each one owns a single mutation, or restructuring computed values into \`\$derived\`. The current pattern works but trips the future-Svelte-warning if the team enables stricter compiler rules.

### 3. \`untrack()\` on initial state reads (lines 33-34)
\`untrack(() => mode === 'loading')\` for initial state init means the state doesn't re-initialize when the \`mode\` prop changes. Intent unclear — either deliberate (component is meant to be re-mounted on mode change) or a bug.

## Why this matters

- AnimationFrames adoption keeps the file consistent with the runed-as-default rule in AGENTS.md
- Effect-mutation pattern is the kind of thing that becomes a compiler error one Svelte version from now
- The untrack pattern is opaque; needs a comment or removal

## See also

- #384, #385, #386, #387, #388, #389 — runed-consistency theme

stickerdaniel:
Closes #390

`src/lib/components/ui/loading-bar/loading-bar.svelte` hand-rolled its render loop with a `rafId` variable, a double-call guard, and three separate `cancelAnimationFrame` sites, even though runed utilities are the repo default. Several `$effect` blocks also mutated `$state` variables (`progressStyle`, `isLoading`), a pattern the Svelte autofixer flags on every block and one that stricter compiler rules will eventually reject, and the `untrack()` calls on the initial state reads were unexplained.

The component now derives `isLoading` from the `mode` prop and computes `progressStyle` with `$derived` from a single `displayWidth` state that only the frame callback writes, so no `$state` is assigned inside any `$effect`. The frame loop is a runed `AnimationFrames` instance created with `immediate: false`, whose internal effect stops the loop on destroy, which makes the `onMount` block and manual cancellation obsolete. `startLoop` guards on `frames.running` because `AnimationFrames.start()` has no double-call guard, and reads it through `untrack` since the spring-sync `$effect` calls it and tracking the running flag would restart the spring animation on every loop start and stop. Mode transitions go through a runed `watch` on `isLoading` that stops the loop when entering the loading state and restarts it when leaving, and the remaining `untrack` around the `initialWidth` capture got a comment marking it as an intentional one-time read.

Verification: `bun scripts/static-checks.ts src/lib/components/ui/loading-bar/loading-bar.svelte` and `bun run test:unit` pass, and the Svelte autofixer reports no remaining state-assigned-in-effect findings on the component. The manual spring and reduced-motion behavior checks on `/shadcn-demo` were not run here.

### DIFF
--- a/src/lib/components/ui/loading-bar/loading-bar.svelte
+++ b/src/lib/components/ui/loading-bar/loading-bar.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
 	import { Progress as ProgressPrimitive } from 'bits-ui';
-	import { onMount, untrack } from 'svelte';
+	import { untrack } from 'svelte';
+	import { AnimationFrames, watch } from 'runed';
 	import { useMotionValue, animate, useReducedMotion } from 'motion-sv';
 	import { getTranslate } from '@tolgee/svelte';
 	import { cn, type WithoutChildrenOrChild } from '$lib/utils.js';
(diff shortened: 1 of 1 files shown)


stickerdaniel:
* **#534** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/534?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/534?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #536, opened 2026-06-11
refactor(chat): Migrate scroll-lock to runed

### LINKED ISSUE #388: Chat scroll-lock: stick-to-bottom-context + chat-container-context duplicate triple-Observer plumbing
## Problem

Two near-identical context classes hand-roll \`ResizeObserver\` + \`MutationObserver\` + \`IntersectionObserver\` lifecycles for the chat scroll-lock feature:

- \`src/lib/components/ai-elements/conversation/stick-to-bottom-context.svelte.ts\`
- \`src/lib/components/prompt-kit/chat-container/chat-container-context.svelte.ts\`

Each owns three observer instances with explicit \`.observe()\` / \`.disconnect()\` calls inside a \`watch()\` cleanup, plus an \`addEventListener('scroll', handler, { passive: true })\` with its matching \`removeEventListener\`.

Runed wraps all three observer types with auto-cleanup, and the \`scroll\` listener fits \`useEventListener\` too.

## Why this matters

1. **Duplication**: two files implement the same scroll-lock pattern. If one fixes a bug (e.g. observer-init race when element re-binds), the other silently regresses.
2. **Cleanup correctness**: each observer needs its own try/catch on \`.disconnect()\` because Three.js / MutationObserver throw if the target has been detached. Runed's wrappers handle that.
3. **Style consistency**: the rest of v2 uses runed for observers.

## Suggested fix

For each file, replace the imperative pattern with:

\`\`\`ts
import { useResizeObserver, useMutationObserver, useIntersectionObserver, useEventListener } from 'runed';

constructor() {
  useResizeObserver(() => this.#element, () => this.#checkScrollPosition());
  useMutationObserver(
    () => this.#element,
    () => { /* recompute */ },
    { childList: true, subtree: true }
  );
  useIntersectionObserver(
    () => this.#sentinel,
    (entries) => { if (entries[0]?.isIntersecting) this.#isAtBottom = true; }
  );
  useEventListener(() => this.#element, 'scroll', this.#onScroll, { passive: true });
}
\`\`\`

And consider whether the two files can collapse to one shared base class — the only difference appears to be the consumer-component naming.

## See also

- #384, #385, #386, #387 — same runed-consistency theme

stickerdaniel:
Closes #388

`src/lib/components/prompt-kit/chat-container/chat-container-context.svelte.ts` and `src/lib/components/ai-elements/conversation/stick-to-bottom-context.svelte.ts` each hand-roll the same scroll-lock plumbing: a `ResizeObserver`, a `MutationObserver`, and an `IntersectionObserver` with explicit `.observe()` and `.disconnect()` calls inside a `watch()` cleanup, plus a manually added and removed `scroll` listener. The rest of the codebase uses runed wrappers for observers, and the duplicated imperative lifecycles mean a fix in one file silently misses the other.

Both classes now register `useEventListener`, `useResizeObserver`, `useMutationObserver`, and `useIntersectionObserver` in the constructor with element getters, so observation and cleanup follow the element binding automatically. The remaining `watch` only owns the sentinel lifecycle, and `#sentinel` became `$state` so the intersection observer re-targets when it changes. `threshold: 0` is passed explicitly because runed defaults to `0.1`, the `characterData` mutation option is kept, and each class keeps its own semantics: 50px bottom threshold plus initial and resize scroll modes in `ChatContainerContext`, 200px threshold and unconditional `#userHasScrolled` reset on intersection in `StickToBottomContext`. The dead `debugInfo` getter in `StickToBottomContext` is removed. Collapsing the two classes into a shared base, as the issue suggests considering, is deferred because they diverge in thresholds, intersection semantics, and scroll modes, so a merge needs a config-object design and a maintainer call.

`bun scripts/static-checks.ts` on both changed files passes, and `bun run test:e2e e2e/ai-chat.spec.ts` passes against the isolated local test stack; community chat scroll behavior was not smoke-tested manually.

### DIFF
--- a/src/lib/components/ai-elements/conversation/stick-to-bottom-context.svelte.ts
+++ b/src/lib/components/ai-elements/conversation/stick-to-bottom-context.svelte.ts
@@ -1,40 +1,86 @@
-import { watch, Context } from 'runed';
+import {
+	watch,
+	Context,
+	useEventListener,
+	useIntersectionObserver,
+	useMutationObserver,
+	useResizeObserver
+} from 'runed';
 
 export class StickToBottomContext {
 	#element: HTMLElement | null = $state(null);
 	#isAtBottom = $state(true);
-	#resizeObserver: ResizeObserver | null = null;
-	#mutationObserver: MutationObserver | null = null;
-	#intersectionObserver: IntersectionObserver | null = null;
-	#sentinel: HTMLElement | null = null;
+	#sentinel: HTMLElement | null = $state(null);
 	#userHasScrolled = $state(false);
 
 	isAtBottom = $derived(this.#isAtBottom);
 
-	// Debug method to help troubleshoot
-	get debugInfo() {
-		if (!this.#element) return null;
-		const { scrollTop, scrollHeight, clientHeight } = this.#element;
-		return {
-			scrollTop,
-			scrollHeight,
-			clientHeight,
-			isAtBottom: this.#isAtBottom,
-			userHasScrolled: this.#userHasScrolled,
-			hasElement: !!this.#element,
-			hasSentinel: !!this.#sentinel
-		};
-	}
-
 	constructor() {
+		// Sentinel lifecycle. Declared before the observers below so the initial
+		// append happens before the MutationObserver starts watching.
 		watch(
 			() => this.#element,
+			(element) => {
+				if (!element) return;
+
+				this.#sentinel = this.#createSentinel(element);
+
+				// Check initial state
+				this.#checkScrollPosition();
+
+				return () => {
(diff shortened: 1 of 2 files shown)


stickerdaniel:
* **#536** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/536?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/536?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #502, opened 2026-06-05
chore: Remove dead demo scaffolding and unused exports

stickerdaniel:
Closes #498
Closes #497

Forks inherited a 1.6 MB Photoshop source in `static/` that shipped to production with every deploy, plus several trees of demo scaffolding with no path into the running app: the `sidebar-07` shadcn demo route with its four backing demo components, the cnblocks `feature-one` demo, the prompt-kit demo conversations, the unregistered `ShadcnDemo` email template, and the unused `ui/chat` primitives with their `use-auto-scroll` hook. Alongside those, a set of alternate-API exports had zero call sites next to live siblings doing the real work (`hapticAction`, the `passkeyQueries` workaround queries, `isRateLimitError`, `parseUserRecord`, `getTemplateNames`, `getTemplateProps`, `translateError`, `createMarketingMarkdownErrorResponse`, `isSvg`), and `ToolBadge` reinvented the Badge primitive with zero importers (the dead-code part of #469; its color-token findings stay open).

Everything listed is deleted. `nav-user.svelte` stays (the live sidebar uses it), and `betterAuth/_generated/api.ts` is updated to match codegen output after the module removal. Two knip findings are intentionally kept and now tagged `@public` with a reason so nobody has to re-investigate them: `optionOrAlt` (keyboard shortcut convention sibling of `cmdOrCtrl`) and `getLegalEmailAddress` (raw mailto variant for #398). The knip ignores that the removal made stale (`src/blocks/**`, `prompt-kit-blocks/**`, `ui/collapsible/**`) are pruned, and the policy is documented in the AGENTS.md regression guard tree. The two oxlint `passkeyQueries` warnings in CI disappear with this change.

`bunx knip` reports no unused files and no unused exports anymore (remaining dependency findings belong to #486). `bun run check:convex` passes, `bun run test:unit` passes (462 tests), `bun scripts/static-checks.ts` passes on all changed files.

See also: #469, #486

PR #538, opened 2026-06-11
fix(a11y): Add accessible names to icon controls

### LINKED ISSUE #454: Icon-only controls lack accessible names; nav-projects trigger hidden from the a11y tree
Several icon-only controls expose no accessible name, and one functional control is removed from the a11y tree:

- [ ] src/lib/components/customer-support/screenshot-editor/ScreenshotToolbar.svelte: nine tool buttons rely on `title` only (not exposed on touch, inconsistent across screen readers)
- [ ] same file, color swatches: `aria-label` comes from hardcoded English color names in screenshot-editor/types.ts:84
- [ ] src/lib/components/authenticated/authenticated-sidebar.svelte:142: the AI-chat-threads chevron `Collapsible.Trigger` has no name (bits-ui adds none)
- [ ] src/routes/[[lang]]/admin/support/thread-chat.svelte:129: skeleton-header back button has no name
- [ ] src/lib/chat/ui/InlineEmailPrompt.svelte:125: unsubscribe button names itself via hover tooltip only
- [ ] src/lib/components/nav-projects.svelte:42: a working `DropdownMenu.Trigger` carries `aria-hidden="true" tabindex={-1}`, so keyboard and screen-reader users can neither reach nor perceive the project actions menu

proposed fix: localized `aria-label` or sr-only text via Tolgee keys, same as data-table-actions.svelte; remove `aria-hidden`/`tabindex` from the functional trigger. The aria-hidden escape hatch is meant for decorative buttons only.

stickerdaniel:
Closes #454

Several icon-only controls expose no accessible name: the nine screenshot toolbar buttons rely on `title` alone (not exposed on touch and inconsistent across screen readers), the color swatches use hardcoded English names from `types.ts`, the AI chat thread chevron `Collapsible.Trigger` in `authenticated-sidebar.svelte` has no name at all, and the skeleton back button in `admin/support/thread-chat.svelte` and the unsubscribe button in `InlineEmailPrompt.svelte` are unnamed for screen reader and touch users.

The toolbar buttons now carry an `aria-label` that reuses each button's existing `title` Tolgee key, so the accessible name always matches the tooltip. `ColorPreset` swaps its hardcoded `name` for a `nameKey` Tolgee key, with six new `support.screenshot.color.*` keys translated in all four locales, and the swatch loop resolves both `title` and `aria-label` through `$t`. The thread chevron gets localized `sr-only` text via a new `aria.toggle_threads` key, following the `data-table-actions.svelte` pattern, the skeleton back button reuses the existing `aria.go_back` key, and the unsubscribe button reuses `chat.email.unsubscribe_tooltip` so its name matches the visible tooltip.

The `nav-projects.svelte` checklist item needs no change because that file was removed in #502. `bun run i18n:push -- --tag-new-keys draft` reported import conflicts and was not forced, so the new keys still need a pull plus conflict resolution before they land in Tolgee Cloud.

`bun scripts/static-checks.ts` on all changed files passes and `bun run test:unit` passes.

### DIFF
--- a/src/i18n/de.json
+++ b/src/i18n/de.json
@@ -398,6 +398,7 @@
 		"target": "Ziel",
 		"toggle_sidebar": "Seitenleiste umschalten",
 		"toggle_theme": "Design umschalten",
+		"toggle_threads": "Unterhaltungsliste umschalten",
 		"view": "Ansicht"
 	},
 	"auth": {
@@ -1055,6 +1056,14 @@
 			},
 			"aria_label": "Screenshot-Editor",
 			"capturing": "Wird aufgenommen...",
+			"color": {
+				"black": "Schwarz",
+				"blue": "Blau",
+				"green": "Grün",
+				"purple": "Lila",
+				"red": "Rot",
+				"yellow": "Gelb"
+			},
 			"next": "Bug Markiert",
 			"tool": {
 				"arrow": "Pfeil (A)",
--- a/src/i18n/en.json
+++ b/src/i18n/en.json
@@ -398,6 +398,7 @@
 		"target": "Target",
 		"toggle_sidebar": "Toggle Sidebar",
 		"toggle_theme": "Toggle theme",
+		"toggle_threads": "Toggle conversation list",
 		"view": "View"
 	},
 	"auth": {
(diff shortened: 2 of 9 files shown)


stickerdaniel:
* **#538** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/538?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/538?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #539, opened 2026-06-11
refactor(sentry): Lazy-load SDK behind DSN gate

### LINKED ISSUE #485: Sentry browser SDK statically imported, ships to first paint when DSN is unset
`src/hooks.client.ts:5` does `import * as Sentry from '@sentry/sveltekit'` as a top-level value import, and `handleError` picks between `Sentry.handleErrorWithSentry(customHandleError)` and `customHandleError` through a runtime ternary on `PUBLIC_SENTRY_DSN` at lines 33-35. Because the namespace value is referenced at runtime, the Sentry browser SDK (one of the heaviest optional client deps) cannot be tree-shaken, so it lands in the client entry that loads on every page including the prerendered marketing landing, even when Sentry is unconfigured. A no-DSN production build ships the full Sentry browser runtime inside the universal client entry.

`AGENTS.md` says static `PUBLIC_*` env enables "dead-code elimination (optional integrations like PostHog are stripped when unconfigured)", and PostHog follows that in `src/lib/analytics/posthog.ts` with `import type PostHog` at line 6 plus `const posthog = (await import('posthog-js')).default` at line 59. Sentry is equally optional and heavier, but uses an unstrippable static namespace import.

Same defect in two more places:

- `src/lib/components/customer-support/customer-support.svelte:15` has the identical top-level `import * as Sentry`, whose only use is the `browser && PUBLIC_SENTRY_DSN ? Sentry.captureException(error) : undefined` ternary at line 180, so the SDK is pulled into the customer-support widget chunk that every interacting user downloads, including on the prerendered home.
- `src/hooks.server.ts:5` is structurally the same (top-level `import * as Sentry`, gated only by runtime ternaries at lines 258 and 269). This is the server hook, so the unstrippable SDK weighs on the server bundle and cold start rather than client first paint.

proposed fix: follow the PostHog pattern. Switch to `import type * as Sentry from '@sentry/sveltekit'`, then `await import('@sentry/sveltekit')` inside the `PUBLIC_SENTRY_DSN` branch to call `init` and build the DSN-only `handleError`/`handle` wrappers, so the SDK is dead-code-eliminated in all three entries when Sentry is unconfigured.

stickerdaniel:
Closes #485

`src/hooks.client.ts`, `src/hooks.server.ts`, and `src/lib/components/customer-support/customer-support.svelte` all did a top-level `import * as Sentry from '@sentry/sveltekit'` and only gated usage with runtime ternaries on `PUBLIC_SENTRY_DSN`. Because the namespace value is referenced at runtime, the SDK cannot be tree-shaken, so a no-DSN build ships the full Sentry browser runtime in the universal client entry that loads on every page, including the prerendered marketing landing, and the same unstrippable import weighs on the server bundle and cold start.

The fix follows the existing PostHog loader pattern. A new `src/lib/monitoring/sentry.ts` exposes a memoized `loadSentry()` that references the SDK only via `import type` plus a dynamic `import('@sentry/sveltekit')` gated on `PUBLIC_SENTRY_DSN`, calls `init` once per process, and emits the existing `devNotice` when the DSN is unset. `hooks.client.ts` now exports a stable `handleError` delegate that upgrades to `handleErrorWithSentry(customHandleError)` once the SDK resolves and falls back to the PostHog path before that, the same race tolerance `getPosthog()` already has. `hooks.server.ts` replaces the spread `sentryHandle()` with a `handleSentry` hook that passes straight through when the DSN is unset and otherwise memoizes a lazily created `sentryHandle()` instance, and `handleError` wraps `handleErrorWithSentry()` the same way while staying `undefined` without a DSN. The customer-support widget opens its capture-error dialog synchronously and fills in the Sentry event id when the lazily loaded SDK resolves. A new banned pattern in `scripts/static-checks.ts` rejects static value imports of `@sentry/sveltekit` under `src/` (`import type` stays allowed) so the regression cannot creep back in.

On a no-DSN production build the server output contains no SDK code at all, and `loadSentry()` compiles down to the `devNotice` plus `return null` in both the client and server graphs, so the client entries (`start`, `app`) carry only the small delegate. Rolldown still emits the SDK chunk as an orphaned asset, but no module in the client graph references it, so it is never fetched. With a dummy DSN set, the build succeeds with the `sentrySvelteKit` plugin active and a small loader chunk lazily imports and initializes the SDK.

`bun scripts/static-checks.ts` on the changed files, `bun run test:unit`, `bun run test:e2e`, and both no-DSN and DSN-set `bun run build` runs pass locally.

### DIFF
--- a/scripts/static-checks.ts
+++ b/scripts/static-checks.ts
@@ -40,7 +40,14 @@ const CONFIG = {
 		deprecated:
 			/ring-offset-background|ring-offset-foreground|text-destructive-foreground|flex-shrink-0|bg-gradient-to-/,
 		/** animate-spin without motion-safe: prefix (WCAG 2.3.3) */
-		bareAnimateSpin: /(?<!motion-safe:)animate-spin/
+		bareAnimateSpin: /(?<!motion-safe:)animate-spin/,
+		/**
+		 * Static value imports/re-exports of the Sentry SDK under src/. They defeat
+		 * dead-code elimination when PUBLIC_SENTRY_DSN is unset and ship the SDK to
+		 * first paint. Lazy-load via $lib/monitoring/sentry instead; `import type`
+		 * stays allowed (erased at build time).
+		 */
+		staticSentryImport: /(?:import|export)\s+(?!type[\s{])[^'"]*from\s*['"]@sentry\/sveltekit['"]/
 	}
 };
 
@@ -231,7 +238,7 @@ async function main(): Promise<void> {
 		}
 		console.log('\n');
 
-		// Banned patterns (deprecated tokens, bare animate-spin)
+		// Banned patterns (deprecated tokens, bare animate-spin, static Sentry imports)
 		printHeader(step++, 'Banned patterns');
 		{
 			const filesToScan = scopedMode
@@ -254,6 +261,11 @@ async function main(): Promise<void> {
 					if (CONFIG.bannedPatterns.bareAnimateSpin.test(line)) {
 						violations.push(
 							`${file}:${i + 1}: bare animate-spin (use motion-safe:animate-spin): ${line.trim()}`
+						);
+					}
+					if (CONFIG.bannedPatterns.staticSentryImport.test(line)) {
+						violations.push(
+							`${file}:${i + 1}: static @sentry/sveltekit import (lazy-load via $lib/monitoring/sentry; import type is allowed): ${line.trim()}`
 						);
 					}
 				}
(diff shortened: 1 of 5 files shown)


stickerdaniel:
* **#539** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/539?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/539?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #542, opened 2026-06-11
fix(chat): Clean up subscription and URL leaks

### LINKED ISSUE #492: Support chatbar leaks its pre-warm Convex subscription and cleanup timeout on unmount
In `src/lib/components/customer-support/ai-chatbar.svelte`, `handleValueChange` opens a live Convex subscription via `client.onUpdate` on the first keystroke to pre-warm the messages query cache, storing the unsubscribe handle in `queryUnsubscribe` (declared at `ai-chatbar.svelte:44`, opened at `ai-chatbar.svelte:121`). The only path that calls `queryUnsubscribe()` is the successful-submit branch, inside a `setTimeout(..., 500)` at `ai-chatbar.svelte:74-79`. The component has no `onDestroy`/`$effect` cleanup, so if it unmounts before submit (user types, then navigates away from the marketing home, unmounting the `LazyCustomerSupport` tree), the WebSocket subscription on the module-level `ConvexClient` singleton leaks along with the pending 500ms timeout. Scope is one idle subscription per session on an anonymous surface, so the impact is small, but it accumulates across navigations.

The same revoke-only-on-explicit-clear shape shows up for object URLs in two places:

- `src/lib/chat/ui/ChatContext.svelte.ts:335` (`uploadFile`) stores an image's `URL.createObjectURL(file)` in `attachment.preview`, and `ChatContext.svelte.ts:456` (`uploadScreenshot`) does the same with `URL.createObjectURL(blob)`. Neither `removeAttachment` (`ChatContext.svelte.ts:252`) nor `clearAttachments` (`ChatContext.svelte.ts:259`) nor any destroy hook ever calls `URL.revokeObjectURL`, so these preview URLs leak on the live chatbar path until the document unloads.
- `src/lib/components/ai-elements/prompt-input/attachments-context.svelte.ts:117` (`add`) creates an object URL per file, revoked only in `remove` (`attachments-context.svelte.ts:129`) and `clear` (`attachments-context.svelte.ts:137`); `PromptInput.svelte` calls `clear()` only on successful submit. This `ai-elements/prompt-input` set is not imported by any route, so it is dead code, but a fork that wires it up copies the leak.

The codebase's own siblings already pair create with a guaranteed revoke, for example the `ChatContext` image-decode helper at `ChatContext.svelte.ts:286-292`, so these are the outliers.

Related: #456 covers a separate dead-code issue in the same widget.

proposed fix: add an `$effect` cleanup (or `onDestroy`) in `ai-chatbar.svelte` that calls `queryUnsubscribe?.()` and clears the pending timeout, and revoke the `preview` object URL in `removeAttachment`/`clearAttachments` plus a destroy-time pass over remaining attachments in `ChatContext.svelte.ts`. For the vendored `attachments-context.svelte.ts`, revoke remaining URLs when the host unmounts.

stickerdaniel:
Closes #492

The support chatbar opens a pre-warm Convex subscription on the first keystroke and only releases it inside a `setTimeout` in the successful-submit branch, so navigating away before submitting leaked both the subscription on the shared `ConvexClient` singleton and the pending timeout. The same release-only-on-success shape applied to attachment previews: `ChatUIContext.uploadFile` and `uploadScreenshot` create `blob:` URLs via `URL.createObjectURL`, and neither `removeAttachment`, `clearAttachments`, upload failure, nor unmount ever revoked them.

`ai-chatbar.svelte` now stores the cleanup timeout id and runs an `onDestroy` that clears the timeout and unsubscribes. `ChatUIContext` gains a `revokePreview` helper and a public `dispose()`: previews are revoked when an attachment is removed, when the composer clears, and when an upload fails, and `dispose()` runs on unmount at the four creator sites (`ChatRoot.svelte` for internally created contexts, `customer-support.svelte`, and both `thread-chat.svelte` routes). To make revocation safe for optimistic rendering, `sanitizeAttachmentsForClone` in `optimistic.ts` now strips the `preview` field so optimistic messages render from the uploaded `url`, which `ChatAttachments.svelte` already falls back to. The vendored `ai-elements/prompt-input` host gets the same unmount guard via `attachments.clear()`.

`bun scripts/static-checks.ts` on the changed files, `bun run test:unit` (incl. new tests for preview stripping and revocation), and `bun run test:e2e -- support-migration.spec.ts` against the isolated local stack all pass.


stickerdaniel:
* **#542** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/542?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/542?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #543, opened 2026-06-11
refactor(convex): Align mutations with conventions

### LINKED ISSUE #496: Convex backend: stock README plus a few convention drifts in mutations and bounded collects
A few spots in the Convex backend drift from the conventions the convex-guidelines skill and `AGENTS.md` mandate. None are live bugs, but every fork inherits them verbatim and the patterns get copied from there.

1. `src/lib/convex/README.md:1-88` is still the verbatim stock Convex scaffold from the initial commit. It shows React `useQuery`/`useMutation` (this project uses `@mmailaender/convex-svelte`), a generic `functions.js`, examples with no `returns` validators (lines 13, 52), and an unbounded `ctx.db.query('tablename').collect()` (line 24) with no `// Bounded:` comment. Each of those contradicts the convex-guidelines skill and the Intentional Anti-Pattern Comments rule in `AGENTS.md`, and it is the first file anyone opening the functions folder reads. Related to the returns-validator work in #461 and the docs-drift cleanup in #474.

2. `updateLastMessage` (`support/threads.ts:655-684`) is an `internalMutation` that three mutation handlers reach via `ctx.runMutation`: `updateThreadHandoff` (`support/threads.ts:462`), support `sendMessage` (`support/messages.ts:110`), and admin `sendAdminReply` (`admin/support/mutations.ts:246`). Convex guidelines reserve `ctx.runMutation` from a mutation for real function boundaries; shared write logic should be a plain helper taking `ctx`, since each `runMutation` opens its own transaction with dispatch overhead and no transactional benefit. Its body uses only `ctx.db.patch` plus two plain helpers, so it is fully extractable. The action call site `createAIResponse` (`support/messages.ts:214`) genuinely needs the wrapper. The same shape recurs across the admin-notification writes: `upsertAdminPreferences` (`admin/notificationPreferences/mutations.ts:151`) and `deactivateAdminPreferences` (`admin/notificationPreferences/mutations.ts:222`) are pure `ctx.db` internalMutations called via `ctx.runMutation` only from mutation-context callers (Better Auth triggers, `setUserRole`, seeders), with no action caller.

3. `upsertSetting` (`admin/founderWelcome/mutations.ts:10`) and `deleteSetting` (`admin/founderWelcome/mutations.ts:33`) read the `adminSettings` singleton via `.first()` on `by_key`, while every reader of the same keys uses `.unique()` (`readFounderWelcomeConfig` in `admin/founderWelcome/queries.ts:44`, `stepDown` in `mutations.ts:115`, `auth.ts` onCreate at line 160). `adminSettings` is a key-value store with one row per key, and `by_key` (`schema.ts:91`) is a non-unique secondary index, so the singleton invariant is enforced only at read level. Using `.first()` in the write path silently keeps a duplicate-key row instead of throwing the way the readers do.

4. The index-scoped `.collect()` in `migrateAnonymousTickets` (`support/migration.ts:39-42`) is bounded to one anonymous user's threads via `by_user` but carries no `// Bounded:` comment, so a reviewer cannot tell at a glance it is intentionally bounded. Same gap on `backfillThreadMetadata` (`aiChat/threads.ts:249`), a full-table `.collect()` on `aiChatThreads` whose twin `support/threads.ts:695` carries the comment, and on `betterAuth/passkeyQueries.ts:35` (`listByUserId`, index-scoped) and `tests.ts:231` (`getSupportThreadsByUserId`, index-scoped, while the same file follows the convention at line 377).

proposed fix: replace `src/lib/convex/README.md` with a short pointer to the convex-guidelines skill (or delete it); extract `syncSupportLastMessage(ctx, threadId)` and `syncAdminPreferences(ctx, args)` helpers and call them directly from the mutation sites, keeping the `internalMutation` wrappers only where an action calls them; switch `upsertSetting`/`deleteSetting` to `.unique()`; and add `// Bounded:` comments to the four index-scoped and migration `.collect()`s.

stickerdaniel:
Closes #496

The Convex backend drifted from the conventions in the `convex-guidelines` skill and `AGENTS.md` in a few spots. Shared write logic in `updateLastMessage`, `upsertAdminPreferences`, and `deactivateAdminPreferences` was reached from mutation handlers via `ctx.runMutation`, which opens a separate transaction with dispatch overhead and no transactional benefit. `upsertSetting` and `deleteSetting` in `admin/founderWelcome/mutations.ts` read the `adminSettings` singleton with `.first()` while every reader of the same keys uses `.unique()`, so a duplicate-key row would be kept silently in the write path instead of throwing. Three index-scoped or migration `.collect()` calls carried no `// Bounded:` comment, so reviewers could not tell at a glance they are intentionally bounded. Every fork inherits these patterns verbatim.

The body of `updateLastMessage` is extracted into an exported plain helper `syncSupportLastMessage(ctx, threadId)` in `support/threads.ts`; the `internalMutation` stays as a thin wrapper because the action `createAIResponse` still needs a real function boundary, while the three mutation-context callers (`updateThreadHandoff`, support `sendMessage`, `sendAdminReply`) now call the helper directly. The bodies of `upsertAdminPreferences` and `deactivateAdminPreferences` move into plain helpers `syncAdminPreferences` and `deactivateAdminPreferencesHelper` in a new `admin/notificationPreferences/helpers.ts` (a separate file avoids the `auth.ts` <- `functions.ts` <- `mutations.ts` import cycle, same pattern as `incrementCounter` in `admin/counters.ts`); both `internalMutation` wrappers are deleted since they had no action callers, and all call sites in `auth.ts`, `admin/mutations.ts`, `localDev.ts`, `previewDev.ts`, and `syncAllAdminPreferences` call the helpers directly. `upsertSetting` and `deleteSetting` now use `.unique()`, matching the readers. `// Bounded:` comments are added to the index-scoped `.collect()` in `migrateAnonymousTickets`, the one-time backfill in `aiChat/threads.ts`, and the test-only `getSupportThreadsByUserId` in `tests.ts`. The stock `README.md` and `passkeyQueries.ts` items from the issue were already resolved on this base and are untouched.

`bun run check:convex`, `bun scripts/static-checks.ts` on all changed files, `bun run test:unit` (487 tests), and `bun run test:e2e e2e/support-migration.spec.ts e2e/admin-settings.spec.ts` all pass.

### DIFF
--- a/src/lib/convex/admin/founderWelcome/mutations.ts
+++ b/src/lib/convex/admin/founderWelcome/mutations.ts
@@ -7,7 +7,7 @@ async function upsertSetting(ctx: MutationCtx, key: string, value: string, admin
 	const existing = await ctx.db
 		.query('adminSettings')
 		.withIndex('by_key', (q) => q.eq('key', key))
-		.first();
+		.unique();
 
 	if (existing) {
 		await ctx.db.patch(existing._id, {
@@ -30,7 +30,7 @@ async function deleteSetting(ctx: MutationCtx, key: string) {
 	const existing = await ctx.db
 		.query('adminSettings')
 		.withIndex('by_key', (q) => q.eq('key', key))
-		.first();
+		.unique();
 
 	if (existing) {
 		await ctx.db.delete(existing._id);
--- a/src/lib/convex/admin/founderWelcome/mutations.ts
+++ b/src/lib/convex/admin/founderWelcome/mutations.ts
@@ -74,7 +74,7 @@ export const updateConfig = adminMutation({
 		// Upsert per-user profile (name, title, replyTo)
 		const existingProfile = await ctx.db
 			.query('adminProfiles')
-			.withIndex('by_userId', (q) => q.eq('userId', adminUserId))
+			.withIndex('by_user', (q) => q.eq('userId', adminUserId))
 			.unique();
 
 		if (existingProfile) {
--- a/src/lib/convex/admin/founderWelcome/queries.ts
+++ b/src/lib/convex/admin/founderWelcome/queries.ts
@@ -63,7 +63,7 @@ async function readFounderWelcomeConfig(ctx: QueryCtx): Promise<FounderWelcomeCo
 	// Read per-user profile for contact person
 	const profile = await ctx.db
 		.query('adminProfiles')
-		.withIndex('by_userId', (q) => q.eq('userId', contactUserId))
+		.withIndex('by_user', (q) => q.eq('userId', contactUserId))
 		.unique();
 
 	// Read global email config
@@ -110,7 +110,7 @@ export const getFounderWelcomeConfig = adminQuery({
 		// Read viewer's own profile for form pre-fill
 		const viewerProfile = await ctx.db
 			.query('adminProfiles')
-			.withIndex('by_userId', (q) => q.eq('userId', ctx.user._id))
+			.withIndex('by_user', (q) => q.eq('userId', ctx.user._id))
 			.unique();
 
 		const viewerUser = await ctx.runQuery(components.betterAuth.adapter.findOne, {
--- a/src/lib/convex/admin/mutations.ts
+++ b/src/lib/convex/admin/mutations.ts
@@ -1,9 +1,13 @@
 import { internalMutation, type MutationCtx } from '../_generated/server';
-import { components, internal } from '../_generated/api';
+import { components } from '../_generated/api';
 import { v, ConvexError } from 'convex/values';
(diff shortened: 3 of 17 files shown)


stickerdaniel:
* **#543** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/543?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/543?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #544, opened 2026-06-11
fix(ui): Handle rejected send and delete callbacks

### LINKED ISSUE #489: Reusable send/delete primitives clear state before awaiting the callback with no rejection handling
Two related client-side error-handling gaps in shared chat and dialog primitives. Both work today because every current consumer happens to guard correctly, but the primitives put the burden on each future consumer, and one that forgets loses data silently.

Send primitives clear the input before the send callback resolves, with no try/catch or rollback inside the primitive:

- `src/lib/chat/ui/ChatInput.svelte:116-124`. `handleSend` calls `ctx.clearInput()`, kicks off `onSend?.(prompt)`, then `ctx.clearAttachments()` and `await`s the promise, with no try/catch, no `onError` prop, and no rollback. If `onSend` rejects, the typed message is already gone and nothing surfaces from the primitive.
- `src/lib/components/customer-support/ai-chatbar.svelte:53`. Sets `input = ''` before awaiting `threadContext.sendMessage(...)`. The catch at `ai-chatbar.svelte:80-97` only toasts and handles rate-limit, never restoring the input.
- `src/routes/[[lang]]/app/community-chat/+page.svelte:131`. Sets `inputValue = ''` before awaiting `client.mutation(api.messages.send, ...)`. The catch at `community-chat/+page.svelte:158-164` only logs and toasts without restoring `inputValue`.

Fire-and-forget mutations use `.finally()` with no `.catch()`, so a transient failure becomes an unhandled promise rejection (`hooks.client.ts` has no global `unhandledrejection` handler):

- `src/routes/[[lang]]/app/+layout.svelte:45-47`. `client.mutation(api.aiChat.threads.getOrCreateWarmThread, {}).finally(...)` resets the in-flight flag with no `.catch()`. The same mutation with the same in-flight-flag pattern at `ai-chat/+page.svelte:48-60` correctly chains `.catch().finally()`.
- `src/lib/components/ui/confirm-delete-dialog/confirm-delete-dialog.svelte:36-43`. `onConfirm().then(...).finally(...)` with no `.catch()`. `onConfirm` is typed `() => Promise<unknown>` (`confirm-delete-dialog.svelte:67`) and callers pass delete mutations, so a rejecting delete is an unhandled rejection while `.finally()` only resets `loading`.
- `src/lib/components/ui/confirm-delete-dialog/confirm-delete-dialog.svelte:73`. The `skipConfirmation` fast path calls `options.onConfirm()` bare, so a rejected delete is also unhandled.

Related to the divergent send implementations in #465.

proposed fix: in `ChatInput.handleSend` capture the input and attachments, clear them, then await `onSend` in a try/catch that restores them on rejection (optionally exposing an `onError` prop), and apply the same capture-and-restore in the two send consumers. For the fire-and-forget mutations, add `.catch()` (matching the `ai-chat/+page.svelte` pattern) before `.finally()`, and wrap the `skipConfirmation` `onConfirm()` call so a rejected delete is logged.

stickerdaniel:
Closes #489

The reusable send and delete primitives cleared state before awaiting their callback and had no rejection handling. `ChatInput.handleSend` called `ctx.clearInput()` and `ctx.clearAttachments()` and then awaited `onSend` with no try/catch, so any consumer whose `onSend` rejects lost the typed message and attachments silently. The support widget's `onSend` catch in `feedback-widget.svelte` toasted without restoring the cleared input. In `confirm-delete-dialog.svelte`, `onConfirm().then().finally()` had no `.catch()` and the `skipConfirmation` fast path called `onConfirm()` bare, so a rejecting delete became an unhandled promise rejection.

`ChatInput.handleSend` now captures the prompt and attachments before clearing, invokes and awaits `onSend` inside a try/catch (so a synchronous throw hits the same rollback as a rejection), and restores both unless the user already typed or attached something new. Restored attachments get their revoked `blob:` previews stripped so thumbnails fall back to the uploaded URL instead of a dead object URL. The `feedback-widget.svelte` catch restores the cleared input the same way `ai-chatbar.svelte` and community chat already do, then rethrows after its toast and rate-limit handling so `ChatInput`'s rollback restores the attachments (the input-restore guard skips the value the widget already put back). `confirm-delete-dialog.svelte` chains a `.catch()` that logs and keeps the dialog open for retry (callers like `recipients-actions.svelte` toast and rethrow), and the `skipConfirmation` path now catches and logs as well.

The input restores in `ai-chatbar.svelte` and `community-chat/+page.svelte` and the `.catch()` on the warm-thread mutation in `app/+layout.svelte` cited in the issue are already present on the base branch, so this PR only covers the remaining primitives.

`bun scripts/static-checks.ts` passes on the three changed files.


stickerdaniel:
* **#544** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/544?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/544?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #546, opened 2026-06-11
fix(support): Toast subscribe and handoff errors

### LINKED ISSUE #487: Support widget email-subscribe and handoff failures revert silently with no user feedback
In the support widget, `setNotificationEmail` and `requestHandoff` catch a failed mutation, roll back the optimistic state, set `this.error` via `setError(...)`, and return `false`, but `error` is never rendered anywhere. The user clicks subscribe or "Talk to support", the optimistic UI flips, the mutation fails, and the UI silently reverts with no feedback.

`support-thread-context.svelte.ts:67` declares `error = $state<string | null>(null)`. `requestHandoff` calls `this.setError(...)` at `support-thread-context.svelte.ts:281` and returns false; `setNotificationEmail` does the same at `support-thread-context.svelte.ts:344`. But `threadContext.error` is read in no template under `src/lib/components/customer-support/`. The only user-facing error path is `toast.error(...)` for send failures in `feedback-widget.svelte:298-313`, so sends get a localized toast while subscribe and handoff get nothing. `handleRequestHandoff` at `feedback-widget.svelte:113-115` discards the false return entirely. For email, `handleSubmitEmail` at `feedback-widget.svelte:118-123` does rethrow on failure, but `InlineEmailPrompt.handleSubmit` swallows it in its catch at `InlineEmailPrompt.svelte:77-79` (and `handleUnsubscribe` likewise at `InlineEmailPrompt.svelte:95-98`), rolling back optimistic state without surfacing anything.

Convex auto-reverts optimistic state on a rejected mutation, so the rollback works, but the app still owes the user an error. Same support-widget context class as #456.

The shared `confirmDelete` primitive has the same silent-failure shape. `confirm-delete-dialog.svelte:36-43` chains `options.onConfirm().then(...).finally(...)` with no `.catch()`, so a rejecting delete leaves the dialog open and surfaces no error, relying entirely on each caller to toast. The `skipConfirmation` fast path at `confirm-delete-dialog.svelte:73` calls `options.onConfirm()` bare (no await, no `.catch()`), so a failed delete becomes an unhandled rejection with no feedback.

proposed fix: render `threadContext.error` in `feedback-widget.svelte` (toast or inline banner) the same way send failures are toasted, and clear it on the next successful action; in `InlineEmailPrompt` surface the caught rejection instead of only rolling back. For the delete primitive, add a `.catch()` to both paths that keeps the dialog open and reports the failure.

stickerdaniel:
Closes #487

In the support widget, a failed email-subscribe or handoff mutation rolled back the optimistic state and set `threadContext.error`, but that error was never rendered anywhere. `handleRequestHandoff` in `feedback-widget.svelte` discarded the `false` return of `requestHandoff`, and `InlineEmailPrompt` swallowed the rejection from `onSubmitEmail` in both `handleSubmit` and `handleUnsubscribe`, so subscribing, unsubscribing, and "Talk to support" all reverted silently while send failures already got a localized toast.

Failures now surface the same way send failures do. `handleRequestHandoff` checks the return value and shows a localized `support.widget.error.handoff_failed` toast with an error haptic. `InlineEmailPrompt` toasts the existing `chat.email.save_failed` and `chat.email.unsubscribe_failed` keys in its catch handlers while still rolling back the optimistic state. The error strings in `SupportThreadContext` stay hardcoded English diagnostics because the context class has no Tolgee access, so the localized user-facing message lives at the call sites; the context now clears its stale `error` on the next successful `requestHandoff` or `setNotificationEmail`. The `handoff_failed` key was added to all four locales, the email keys already existed everywhere.

The `confirm-delete-dialog` part of the issue is already fixed on the base branch by #489, which added `.catch()` to both confirm paths. The new key has been pushed to Tolgee with `--force-mode KEEP` and verified present in all four languages via a fresh pull.

`bun scripts/static-checks.ts` on the changed files passes, no Convex files changed.


stickerdaniel:
* **#546** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/546?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/546?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #551, opened 2026-06-12
chore(eslint): Add retro regression guard rules

stickerdaniel:
See also: #508, #500, #475

Three anti-patterns were fixed across the recent fix PRs, and all three issues explicitly deferred a lint rule as follow-up. Nothing currently prevents reintroduction: a copy-pasted bare `test.skip()` reintroduces the timing-race dodge, a new module-scope class instance in a `.svelte.ts` file reintroduces the cross-request SSR state leak, and a new `fly`/`slide` transition with literal params silently ignores `prefers-reduced-motion` again.

This adds three custom ESLint rules following the existing `eslint/rules/` pattern, each with a co-located vitest suite. `local/no-bare-test-skip` (scoped to `e2e/**/*.spec.ts`) flags `test.skip()` with zero arguments and points to auto-waiting assertions or `test.skip(condition, 'reason')` gating, the style already used in `e2e/signin-last-used-badge.spec.ts`. `local/no-module-state-singleton` (scoped to `**/*.svelte.ts`) flags top-level `export const x = new SomeClass(...)` while exempting runed `Context`, the sanctioned per-request sharing pattern. `local/require-motion-guard-transition` (scoped to `**/*.svelte`) flags `fly` and `slide` directives (`transition:`, `in:`, `out:`) that have no params or only plain literal params; any conditional, identifier, member or call value counts as guarded, so both sanctioned styles (`duration: skipTransition ? 0 : 200` in `sliding-header.svelte` and `y: prefersReducedMotion.current ? 0 : 10` in `ConversationScrollButton.svelte`) pass unchanged, and `fade` stays exempt as opacity-only.

The two intentional singletons named in #500 got an `eslint-disable-next-line` with a short reason next to their existing rationale comments: `haptic` in `use-haptic.svelte.ts` (browser-guarded writes, SSR-safe) and `supportUserId` in `support-user-id.svelte.ts` (`PersistedState`, browser-written only). The repo sweep surfaced a third module with the same pattern, `last-auth-method.svelte.ts`, whose two `PersistedState` exports got the same disable. The motion rule checks guards syntactically; data-flow analysis that would verify an identifier param object actually zeroes motion was out of scope.

Full `eslint .` run passes with zero findings, `bun scripts/static-checks.ts` on all changed files passes, and `bun run test:unit` passes (563 tests, including the 27 new rule tests).

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -15,6 +15,9 @@ import noHardcodedSrOnlyRule from './eslint/rules/no-hardcoded-sr-only.js';
 import noDebounceInRuneRule from './eslint/rules/no-debounce-in-rune.js';
 import noHardcodedModifierKeysRule from './eslint/rules/no-hardcoded-modifier-keys.js';
 import requireReturnsValidatorRule from './eslint/rules/require-returns-validator.js';
+import noBareTestSkipRule from './eslint/rules/no-bare-test-skip.js';
+import noModuleStateSingletonRule from './eslint/rules/no-module-state-singleton.js';
+import requireMotionGuardTransitionRule from './eslint/rules/require-motion-guard-transition.js';
 
 const gitignorePath = path.resolve(import.meta.dirname, '.gitignore');
 const localPlugin = {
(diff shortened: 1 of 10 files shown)


stickerdaniel:
* **#551** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/551?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/551?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #552, opened 2026-06-12
test(i18n): Guard literal key references against en.json

stickerdaniel:
During the recent fix wave the only check that a translation key referenced in code actually exists in `src/i18n/en.json` was manual review. A missing key fails silently: the Tolgee `<T>` component, `$t()` and the Convex `t()` helper all fall back to the key itself, so production renders raw strings like `common.error` instead of copy.

`scripts/i18n-used-keys.test.ts` walks `src/**/*.{svelte,ts}` with a plain sorted fs walk (skipping `_generated`, `src/i18n` and test files) and extracts every literal key reference: `keyName="..."` and `keyName={'...'}` attributes, the first string argument of `$t(...)`, and the second argument of the Convex `t(locale, key)` helper (scoped to `src/lib/convex/`). Each key must resolve to a string leaf in `en.json`, and failures list every missing key with its `file:line`. Frontend references are checked against Tolgee's flattening semantics, since `en.json` contains flat dotted keys like `"signup.title"` inside `auth`, while Convex references use the strict nested resolution of `getNestedValue()`. Dynamic references (identifiers, ternaries, template literals and Svelte-interpolated attributes like `keyName="pricing.features.free.{key}"`) cannot be resolved statically and are skipped, documented in the test. `locale-parity.test.ts` already guarantees `de`/`es`/`fr` mirror `en.json`, so checking `en.json` covers all locales.

The first run surfaced three real gaps. `common.error` (sign-out error toast in `nav-user.svelte`) was missing from every locale file and would have rendered the raw key, so it is now added to all four locales. `auth.signin.oauth_last_used` was missing everywhere and only masked in English by its `defaultValue="Last used"`, leaving German, Spanish and French users with the English fallback; it is now translated in all four locales. The admin support reply error toast referenced `admin.support.chat.error.send_failed`, a key that never existed, while the intended `admin.support.chat.send_error` sat orphaned in all locale files; the code now uses the existing key. Four `auth.signup.*` hits from the naive first pass turned out to be valid flat dotted keys, which is what motivated the Tolgee flattening semantics above. The new keys are pushed to Tolgee tagged `draft`.

`bun run test:unit` passes including the new guard. `bun scripts/static-checks.ts` passes on the changed files.

### DIFF
--- a/scripts/i18n-used-keys.test.ts
+++ b/scripts/i18n-used-keys.test.ts
@@ -1,0 +1,165 @@
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+import { describe, expect, it } from 'vitest';
+import en from '../src/i18n/en.json';
+
+// Every translation key referenced as a string literal in code must exist in
+// en.json — a missing key renders as the raw key string in production (the
+// Tolgee <T> component, $t() and the Convex t() helper all fall back to the
+// key itself). locale-parity.test.ts guarantees de/es/fr mirror en.json, so
+// checking against en.json covers all locales.
+//
+// LIMITATION: this is regex-level extraction of LITERAL references only.
+// Dynamic keys are silently skipped because they cannot be resolved
+// statically: identifiers (keyName={item.translationKey}), ternaries
+// (keyName={cond ? 'a' : 'b'}), template literals ($t(`a.${x}`)), and
+// Svelte-interpolated attributes (keyName="pricing.features.free.{key}").
+
+const SCAN_ROOT = path.resolve('src');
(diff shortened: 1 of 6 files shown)


stickerdaniel:
* **#552** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/552?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/552?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Issue #483, opened 2026-06-04
CI workflows: Graphite token passed on the command line and missing least-privilege permissions


stickerdaniel:
Three secret-hygiene and least-privilege gaps in `.github/workflows`, fixable in one consistency pass.

`graphite-pr-sync.yml:31` runs `gt auth --token ${{ secrets.GRAPHITE_TOKEN }}`, interpolating the secret into the run command line. That puts it in the process arg list (readable via `ps`, captured by security audit events) and in the rendered script. The other workflows reference secrets through `env:` or step inputs instead (`e2e-fork-preview.yml`, `e2e-preview-cf.yml`, `e2e-preview-vercel.yml`, `claude.yml`), so the command line stays clean. GitHub's "Using secrets" doc recommends keeping secrets off the command line for this reason. Runners here are ephemeral single-tenant VMs, so this is audit-surface and convention drift, with no active exposure.

`static-checks.yml` declares no top-level `permissions:` and only `e2e-pending` (44-45) sets `statuses: write`. The `lint` (9-23), `typecheck` (25-35), and `unit-tests` (56-66) jobs inherit the default `GITHUB_TOKEN` scopes and none use the token, so each needs only `contents: read`. OpenSSF Scorecard Token-Permissions docks a point when the top level is unscoped even if every job is scoped.

`e2e-preview-vercel.yml` has no `permissions:` block at all. The `e2e-tests` job (7-21) fetches config and runs Playwright, posts no commit status, and needs only `contents: read`. Every other workflow here declares least-privilege already: `e2e-preview-cf.yml:12-13`, `e2e-fork-preview.yml:33-35`, and the `e2e-pending` job in `static-checks.yml`.

proposed fix: in `graphite-pr-sync.yml` set `env: { GRAPHITE_TOKEN: ${{ secrets.GRAPHITE_TOKEN }} }` on the auth step and call `gt auth --token "$GRAPHITE_TOKEN"`. Add a single top-level `permissions: { contents: read }` to `static-checks.yml` (keeping the `e2e-pending` `statuses: write` override) and to `e2e-preview-vercel.yml`.

PR #553, opened 2026-06-12
ci(guards): Add actionlint and batch-merge guards

stickerdaniel:
See also: #481, #483, #473

The recent 40-PR merge wave exposed process gaps that no automated guard covered: a shell-injection class in CI workflows (#481), workflow permission gaps (#483), shell-string spawning in scripts (#473), and a batch-merge incident where two individually green PRs broke main semantically (the same function landed from two PRs, and one PR removed a schema index while another added code querying it; hotfixed in #550).

This adds `.github/workflows/actionlint.yml`, which lints all workflows on `pull_request` and on push to main (path-filtered to `.github/workflows/**`). It uses `reviewdog/action-actionlint` pinned to the verified v1.72.0 commit SHA with `fail_level: error`, so findings fail the job instead of only annotating, and the `github-annotations` reporter keeps the job on read-only permissions. A new `.github/actionlint.yaml` declares the `ubicloud-standard-2` runner label. Making the check green from day one meant fixing what it surfaced in the two e2e preview workflows: the `SC2129` grouped-redirect findings (which the action maps to error severity), quoting around `$CONFIG` and `$GITHUB_ENV`, and an explicit `shellcheck disable=SC2086` with a reason on the intentionally word-split curl header variables.

`scripts/static-checks.ts` gains a banned pattern for `execSync(` so the shell-string spawn class removed in #473/#514 cannot reappear in `src/`; the repo currently has zero hits, and spawn-style argument arrays (`runCommandCapture` in `scripts/deploy/utils.ts`) remain the sanctioned path. `.gitignore` adds `.worktree-ready` so worktree provisioning markers never get committed. AGENTS.md gets a new Regression Guard Decision Tree entry for the batch-merge incident class that #550 hotfixed, reflecting that `required_status_checks.strict` is now enabled on main, plus a Convex Backend note about reverting local `_generated` churn before committing.

The Regression Guard Decision Tree itself is now a hard contract instead of advice, with the obligation on the decision rather than the guard: every fix PR records one of three verdicts (`Regression guard: added <name>`, `covered by <name>`, or `not warranted, <reason>`) directly above its verification line, and the new `require-regression-guard.yml` workflow fails fix-titled PRs from humans that omit the line (bots like Renovate are exempt; the PR body is passed via `env:` so the check is not an injection sink). Guards are scoped to recurring bug classes (multiple occurrences, convention copy-paste surface, silent failures); singular logic errors take the `not warranted` verdict instead of accumulating one-off instance tests in CI.

`bun scripts/static-checks.ts` passes on all changed files, `actionlint` reports zero findings across `.github/workflows`, the new banned pattern fires on a test fixture, and both new YAML files parse with pyyaml.

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,0 +1,4 @@
+self-hosted-runner:
+  # Ubicloud managed runners used by all workflows in this repo
+  labels:
+    - ubicloud-standard-2
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -1,0 +1,30 @@
+name: Actionlint
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - '.github/workflows/**'
+      - '.github/actionlint.yaml'
+  pull_request:
+    paths:
+      - '.github/workflows/**'
+      - '.github/actionlint.yaml'
+
+permissions:
+  contents: read
+
+jobs:
+  actionlint:
+    name: Lint Workflows
+    runs-on: ubicloud-standard-2
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+      - uses: reviewdog/action-actionlint@6fb7acc99f4a1008869fa8a0f09cfca740837d9d # v1.72.0
+        with:
+          # Log-only reporter: needs no write permissions and works on both
+          # pull_request and push events. fail_level makes findings fail the job
+          # instead of only annotating.
+          reporter: github-annotations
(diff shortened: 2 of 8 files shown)


stickerdaniel:
* **#553** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/553?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/553?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Issue #466, opened 2026-06-04
Message-quota banner duplicated with diverged thresholds and key namespaces


stickerdaniel:
community-chat/+page.svelte:282 and ai-chat/thread-chat.svelte:162 render a visually identical out-of-messages / low-messages banner, with two parallel translation namespaces (`chat.alerts.*` vs `ai_chat.alerts.*`) and two different low-message thresholds (absolute `<= 3` vs ratio `<= 1/3`). A user crossing features gets inconsistent warnings, and every key exists twice in all four locales.

proposed fix: extract one quota-banner component taking `(isPro, remaining, total, onUpgrade, isUpgrading)` with one shared key set and one threshold rule.

PR #554, opened 2026-06-12
chore(i18n): Enforce nested key structure

stickerdaniel:
See also: #552, #466

The locale files mixed two structures: 859 keys as nested objects and a handful as flat dotted key names (`auth.signup.title` literally stored as a dotted string key). The dotted entries were not a style choice; they were Tolgee's export fallback for two key collisions where a string leaf also served as a namespace prefix: `auth.signup` ("Sign Up", referenced nowhere in code) blocked nesting of the five `auth.signup.*` keys, and the stale pre-#466 keys `chat.alerts.low_messages.{title,description,description_plural}` collided with the live `chat.alerts.low_messages` leaf the quota banner uses. The mix forced consumers like the new `i18n-used-keys` guard in #552 to support two resolution semantics.

The convention is now nested only. The four colliding keys were deleted in Tolgee (verified: a fresh `tolgee pull` exports zero dotted leaves, `auth.signup` as a proper object, and the live `low_messages` leaf intact), `auth.signup.*` is a nested group in all four locale files, and `scripts/locale-parity.test.ts` gains a structure test that rejects any dotted key name at any depth, so the collision class cannot be committed again. AGENTS.md documents the rule next to the existing translation guidance.

`bun run test:unit` passes (540 tests, including the 4 new structure tests) and `bun scripts/static-checks.ts` passes on all changed files.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -170,6 +170,7 @@ These commands use `varlock run` to load env vars from `.env.schema` + `.env.loc
 
 - `bun run i18n:pull` - Download latest translations from Tolgee Cloud. Run this ALWAYS before making any changes to the `src/i18n/*` json translation files.
 - When adding new translation keys, ALWAYS add translations for ALL languages in the `src/i18n/*` json translation files.
+- Key names never contain literal dots; nesting is the only structure. A dotted leaf means a key in Tolgee is both a string leaf and a namespace prefix; resolve the collision in Tolgee instead of committing the dotted fallback (enforced by `scripts/locale-parity.test.ts`).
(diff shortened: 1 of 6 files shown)


stickerdaniel:
* **#554** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/554?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/554?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #556, opened 2026-06-12
fix(ui): Move sidebar badge into shortcut overlay

### LINKED ISSUE #555: Admin sidebar support badge overlaps the shortcut hint and ignores the pressed state
The Support item in the admin sidebar shows a badge with the number of support threads awaiting a response. Hovering the item also reveals its keyboard shortcut hint (⌃⇧3). Both are pinned to the right edge of the row, so they render on top of each other: the hint is absolutely positioned inside the menu button (src/lib/components/authenticated/authenticated-sidebar.svelte:46) and the badge is absolutely positioned as a sibling of the button via Sidebar.MenuBadge (src/lib/components/authenticated/authenticated-sidebar.svelte:181). As soon as the hint fades in on hover, it collides with the badge.

The badge also doesn't follow the pressed state. The menu button translates down by 1px while pressed (active:translate-y-px in src/lib/components/ui/sidebar/sidebar-menu-button.svelte:5). The shortcut hint moves with it because it lives inside the button, but the badge is a sibling of the button inside the menu item, so it stays in place and visually detaches from the row while clicking.

proposed fix: render the badge inside the same right-edge overlay strip as the shortcut hint. The badge then sits in normal flow at the right end of the strip, the kbd hint automatically shifts left by exactly the badge's width whenever a badge is present (any digit count, including the 99+ cap), and both inherit the button's press translation and vertical centering. The strip's opaque hover background switches from an opacity toggle on the whole strip to a background toggle, so the badge stays visible at rest while the kbd part still only appears on hover.

```diff
@@ src/lib/components/authenticated/authenticated-sidebar.svelte (overlay snippet)
-{#snippet shortcutHint(keys: string[])}
-	<span class="... bg-sidebar-accent-hover ... opacity-0 group-hover/menu-button:opacity-100 ...">
-		<Kbd.Group>...</Kbd.Group>
-	</span>
-{/snippet}
+{#snippet trailingOverlay(keys: string[] | undefined, badge: number | undefined)}
+	<span class="... group-hover/menu-button:bg-sidebar-accent-hover ...">
+		{#if keys}
+			<Kbd.Group class="opacity-0 group-hover/menu-button:opacity-100">...</Kbd.Group>
+		{/if}
+		{#if badge && badge > 0}
+			<span class="...">{badge >= 100 ? '99+' : badge}</span>
+		{/if}
+	</span>
+{/snippet}
@@ src/lib/components/authenticated/authenticated-sidebar.svelte (menu item)
-	{#if item.badge && item.badge > 0}
-		<Sidebar.MenuBadge>{item.badge >= 100 ? '99+' : item.badge}</Sidebar.MenuBadge>
-	{/if}
```

stickerdaniel:
Closes #555

The admin sidebar renders the support count badge and the keyboard shortcut hint pinned to the same right edge of the nav row. Hovering a row with a badge faded the shortcut hint in directly on top of the count. The badge was also a sibling of the menu button instead of a child, so it ignored the button's 1px pressed-state translation and visually detached from the row while clicking.

The badge now renders inside the same right-edge overlay as the shortcut hint. Both sit in normal flow within the absolutely positioned strip, so a visible badge pushes the hint left by exactly its own width (any count, including the 99+ cap), and both inherit the button's press translation and vertical centering. The strip's opaque hover background moved to a `::before` layer that fades in below the content, so the badge stays visible at rest while the kbd hint still only appears on hover. The now unused `Sidebar.MenuBadge` usage was removed.

`bun scripts/static-checks.ts src/lib/components/authenticated/authenticated-sidebar.svelte` passes. Verified locally against a seeded awaiting support thread in dark and light mode: badge at rest, hover with the hint left of the badge, active route styling, and press translation.

### DIFF
--- a/src/lib/components/authenticated/authenticated-sidebar.svelte
+++ b/src/lib/components/authenticated/authenticated-sidebar.svelte
@@ -31,27 +31,42 @@
 		<item.icon />
 	{/if}
 	<span class="min-w-0 truncate"><T keyName={item.translationKey} /></span>
-	{#if item.kbd}
-		{@render shortcutHint(item.kbd)}
+	{#if item.kbd || (item.badge ?? 0) > 0}
+		{@render trailingOverlay(item.kbd, item.badge)}
 	{/if}
 {/snippet}
 
-<!-- Shortcut hint, revealed on hover. Absolutely placed so it never steals width
-	 from the label when hidden. The background is the opaque match for the row's
-	 hover colour (sidebar-accent-hover), switching to sidebar-accent only once the
-	 row is the active route. Pressing keeps the hover colour to match the row, which
-	 stays on its hover background until release. A left mask fades that edge into the
-	 label, pure alpha so it never muddies the colour. pr-8 makes room for the
+<!-- Right-edge overlay holding the optional count badge and the shortcut hint that is
+	 revealed on hover. Absolutely placed so it never steals width from the label when
+	 hidden. The badge sits at the strip's right end in normal flow, so a visible badge
+	 pushes the shortcut hint left by exactly its own width, and both live inside the
+	 menu button so they ride its pressed-state translation. The hover background lives
+	 on a ::before layer (kept below the content by the stacking context the mask
+	 creates) so it can fade in without hiding the badge: the opaque match for the
+	 row's hover colour (sidebar-accent-hover), switching to sidebar-accent only once
+	 the row is the active route. Pressing keeps the hover colour to match the row,
+	 which stays on its hover background until release. A left mask fades that edge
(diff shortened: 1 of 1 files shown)


stickerdaniel:
* **#556** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/556?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/556?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #557, opened 2026-06-12
chore(deploy): Drop build-env preview password override

stickerdaniel:
The preview admin password had two competing sources: the Convex project default env var for preview deployments (documented as the canonical place) and a `PREVIEW_ADMIN_PASSWORD` in the hosting platform build env, which `scripts/deploy/steps.ts` wrote into every preview deployment via `convex env set`, silently overriding the default. A stale value in the CF Workers Builds env did exactly that, so logging in with the documented default password failed and nothing pointed at the cause.

Remove the override path from the deploy script and the corresponding row from the README's hosting platform env table. The Convex preview default (`bunx convex env default set --type preview PREVIEW_ADMIN_PASSWORD`) is now the single source, and the hosting build env only carries build concerns like deploy keys. The seeding failure warning no longer suggests the build env as an alternative.

`bun scripts/static-checks.ts scripts/deploy/steps.ts README.md` passes. The stale `PREVIEW_ADMIN_PASSWORD` build var in CF Workers Builds gets removed alongside this change so future previews seed with the project default.

### DIFF
--- a/.env-convex.schema
+++ b/.env-convex.schema
@@ -85,7 +85,8 @@ AUTH_E2E_TEST_SECRET=
 # ====================================
 
 # Password for the auto-seeded preview admin (admin@preview.dev)
-# Only set on preview Convex deployments by deploy.ts
+# Inherited by preview deployments from the Convex project preview default
+# env var: bunx convex env default set --type preview PREVIEW_ADMIN_PASSWORD
 # @optional @sensitive
 PREVIEW_ADMIN_PASSWORD=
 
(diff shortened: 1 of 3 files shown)


stickerdaniel:
* **#557** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/557?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/557?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #558, opened 2026-06-12
fix(dev): Ignore static SITE_URL in .env.convex.local

stickerdaniel:
A `SITE_URL` copied into `.env.convex.local` (from the cloud setup steps, the example file, or because the schema marks it `@required`) overrode the URL that local dev derives from the running Vite port. That pinned Better Auth's trusted origin to the wrong origin and silently broke local admin sign-in, with nothing pointing at the env file as the cause. Until now only doc warnings in AGENTS.md and the example file guarded against this.

`vite.config.ts` now strips `SITE_URL` from the local override file and always uses the derived URL (or `PORTLESS_SITE_URL`, unchanged). When a stripped value differs from the derived one, it logs a warning naming both values and the fix. Cloud and preview deployments are unaffected: `scripts/deploy.ts` sets `SITE_URL` per deployment on Cloudflare and Vercel and never reads `.env.convex.local`, which is gitignored and absent in CI. The stale doc warnings in AGENTS.md and `.env.convex.example` now describe the new behavior.

`bun scripts/static-checks.ts vite.config.ts AGENTS.md .env.convex.example` passes. Verified locally: with `SITE_URL=https://production.example.com` in `.env.convex.local`, `bun run dev` logs the ignore warning and sets `SITE_URL=http://localhost:20191/` on the local backend.

### DIFF
--- a/.env.convex.example
+++ b/.env.convex.example
@@ -77,8 +77,8 @@ AUTH_GITHUB_SECRET=
 # Auto-derived from VERCEL_URL for preview deployments
 # Note: For email images, use EMAIL_ASSET_URL (always production)
 # Local dev: auto-derived from the running Vite port (per-project, see
-# scripts/dev-ports.ts). Leave SITE_URL unset locally so a copied value can't pin
-# the trusted origin to a wrong port and break the admin sign-in.
+# scripts/dev-ports.ts). A value set here is ignored with a warning, because a
+# static URL would pin the trusted origin to a wrong origin and break sign-in.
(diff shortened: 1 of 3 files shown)


stickerdaniel:
* **#558** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/558?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/558?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #559, opened 2026-06-13
refactor(i18n): collapse used-key resolution to one check

stickerdaniel:
See also: #552, #554

The used-keys test carried two resolution semantics: a Tolgee flatten that joins nested keys with dots, and a strict Convex nested walk. The split existed because en.json once held flat dotted key names (a key that is both a string leaf and a namespace prefix), which Tolgee flattens to a valid full key but the Convex walker would miss.

Since #554 added the locale key structure guard in locale-parity.test.ts, dotted key names are banned at every depth, so en.json is purely nested and both runtimes resolve identically. This drops the per-reference Semantics tag and the resolvesToStringLeaf walker, resolving every reference against the single flattened leaf-key set. Extraction stays three-pronged (keyName, $t, and the Convex t() helper each find their own references); only the resolution collapses.

bun run test:unit -- scripts/i18n-used-keys.test.ts passes (2/2). static-checks on the file is clean.

### DIFF
--- a/scripts/i18n-used-keys.test.ts
+++ b/scripts/i18n-used-keys.test.ts
@@ -37,21 +37,19 @@ const TOLGEE_T = /\$t\(\s*(?:'([^'\n]*)'|"([^"\n]*)")/g;
 const CONVEX_T =
 	/(?<![\w$.])t\(\s*(?:['"][a-z]{2}['"]|[\w.$]+(?:\([^()]*\))?)\s*,\s*(?:'([^'\n]*)'|"([^"\n]*)")/g;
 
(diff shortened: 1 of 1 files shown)


stickerdaniel:
* **#559** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/559?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/559?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #562, opened 2026-06-14
Purge Cloudflare edge cache after production deploys

### LINKED ISSUE #561: Purge Cloudflare edge cache after production deploys
Marketing routes that aren't prerendered get edge-cached by the `handleCacheControl` hook: unauthenticated requests matching `matchPublicMarketingRoute()` are served with `Cache-Control: public, s-maxage=3600, stale-while-revalidate=86400` (`src/hooks.server.ts`). Cloudflare then holds the rendered HTML at the edge for up to an hour, and can keep serving it stale for a day while it revalidates.

The production deploy command is a plain `bunx wrangler deploy`, which uploads the new Worker version but never invalidates the edge cache. So after shipping a change to a cached marketing page, visitors keep getting the old HTML until the TTL expires. This only surfaces once a custom domain with edge caching is attached; workers.dev previews aren't affected the same way.

`AGENTS.md` already flags this as a known gap (the cache-purge note near line 527), but it points at `scripts/cf-deploy.ts`, which is the non-production deploy command (`wrangler versions upload --preview-alias`). Purging belongs after the production `wrangler deploy`, not in the preview path.

Proposal: add a `scripts/cf-prod-deploy.ts` wrapper and set it as the production deploy command. It runs `wrangler deploy` and, on success, calls `POST /zones/{zone_id}/purge_cache` with `{"purge_everything": true}`. Gate the purge behind `CF_PURGE_TOKEN` and `CF_ZONE_ID` so it stays a clean no-op for forks without a custom domain. Add both to `.env.schema` (`@sensitive @optional` and `@optional`), and update the AGENTS.md note to point at the new prod wrapper.

stickerdaniel:
Closes #561

Marketing routes that aren't prerendered are edge-cached by the `handleCacheControl` hook (`s-maxage=3600`), but the production deploy command was a plain `wrangler deploy` that never invalidated the cache, so a deploy served stale HTML until the TTL expired. This only surfaces once a custom domain with edge caching is attached.

This adds `scripts/cf-prod-deploy.ts` as the production deploy command. It runs `wrangler deploy` and, on success, purges the zone cache via the Cloudflare API. The purge is gated on `CF_PURGE_TOKEN` and `CF_ZONE_ID`, so forks without a custom domain keep plain `wrangler deploy` behavior. The purge is fail-open: both HTTP errors and network-level throws (DNS, connection refused, timeout) log loudly but exit 0, since the worker is already live and the only consequence is the pre-existing staleness window. Failing the build there would only trigger a confusing retry of an already-published deploy.

Adds `CF_PURGE_TOKEN` and `CF_ZONE_ID` to `.env.schema`, points the AGENTS.md cache-purge note and the README CF setup (including the manual-deploy snippet) at the new wrapper, and documents both vars in the env matrix and the custom-domain section.

Regression guard: not warranted, one-off deploy-command wiring with no recurring code pattern to guard.

`bun scripts/static-checks.ts` passes.

### DIFF
--- a/.env.schema
+++ b/.env.schema
@@ -220,6 +220,16 @@ CF_PAGES_BRANCH=
 # @public @optional
 PRODUCTION_BRANCH=
 
+# Cloudflare API token with Zone > Cache Purge, used by scripts/cf-prod-deploy.ts
+# to purge the edge cache after a production deploy. Purge is skipped when unset.
+# @sensitive @optional
+CF_PURGE_TOKEN=
+
+# Cloudflare zone id for the production custom domain, used by the post-deploy
+# cache purge. Purge is skipped when unset.
+# @public @optional
+CF_ZONE_ID=
+
 # ====================================
 # CI
 # ====================================
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -524,11 +524,11 @@ Markdown responses on the same URLs use `Cache-Control: private` to stay out of 
 
 #### Cache purging (production)
 
-When a custom domain is added to CF Workers, add cache purging to the deploy flow (`scripts/cf-deploy.ts`) to prevent stale responses after production deploys:
+`scripts/cf-prod-deploy.ts` is the production deploy command: it runs `wrangler deploy` and then purges the edge cache, so a new deploy is served immediately instead of stale cached HTML. The purge is a no-op unless both `CF_PURGE_TOKEN` and `CF_ZONE_ID` are set, so forks without a custom domain keep plain `wrangler deploy` behaviour.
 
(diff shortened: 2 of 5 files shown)


stickerdaniel:
* **#562** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/562?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/562?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #563, opened 2026-06-15
Skip Convex env validation when deploy key can't list vars

stickerdaniel:
Convex deploy keys can deploy but are denied `ViewEnvironmentVariables`, so `scripts/validate-convex-env.ts` (which runs `convex env list`) fails in CI whenever the build authenticates with a deployment-scoped deploy key. A leaked deploy key must not be able to read secret values, so this is by design on Convex's side. The pre-flight check then hard-fails the production build.

This skips the validation when `convex env list` fails with a permission error, logging the required var names instead of failing the build. The deploy still runs, and a genuinely missing var surfaces at runtime. Validation still works wherever the auth can read env vars (local dev, project-level keys).

`bun scripts/static-checks.ts scripts/validate-convex-env.ts` passes.


stickerdaniel:
* **#563** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/563?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/563?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #567, opened 2026-06-15
chore(audit): fix consistency, convention and doc findings

stickerdaniel:
A multi-pass audit of the template surfaced 38 verified issues: documentation pointing at files and commands that no longer exist, two regression guards that passed vacuously, duplicated Convex upload and metadata code, dead code left behind by an incorrect doc, and hardcoded strings that bypassed i18n. None were functional or security bugs, but several actively misled maintainers, and one stale doc had even produced unreachable code.

This cleans them up across the stack. The setup guides and AGENTS.md now reference the real `.tolgeerc`, the correct layout paths, and the actual PostHog ad-block check, and the Convex guidelines skill documents that `getAuthUser` throws rather than returning null. The unreachable null guards that the mistaken contract created are removed, and the duplicated upload constraints, blob validation, and file-metadata helpers move into shared `files/upload.ts` and `files/metadata.ts` modules so aiChat and support stop copying them and aiChat no longer reaches into a support internal mutation. The 2000-char message limit becomes a single constant per runtime with a localized error, the chat upload toasts and team-member alt text are localized, the app home gets its own meta keys, `listUsers` returns the `{ items }` shape the table kit expects, and the `RESEND_API_KEY` and `AUTH_EMAIL` schema descriptions are corrected to reflect that the Convex backend consumes them.

The prerender language guard previously compared hardcoded duplicate arrays and never read `svelte.config.js` or `+layout.svelte`, so a wrong language edit passed silently; it now reads the real sources. The marketing-route registration rule gains the sibling test every other rule already had.

Regression guard: added. prerender-sync now reads its real sources, a new test covers the marketing-route rule, and the MIME-parity guard was adapted to the shared upload module. All three were verified to go red against injected regressions before landing.

Verification: `bun run check:convex`, `bun run test:unit` (576 tests), and `bun scripts/static-checks.ts` on every changed file pass locally. E2E runs against the CF preview deploy.


stickerdaniel:
* **#567** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/567?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/567?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #568, opened 2026-06-15
fix: align sidebar, select, and password-meter to their surfaces

stickerdaniel:
Three small surface/theme issues that are easy to miss with the default zinc theme but show up in forks that customise the background or card colours.

The select dropdown rendered its rounded items flush against the container edge because the viewport had no padding, unlike the dropdown-menu which insets its items. The select viewport now uses matching padding so both dropdowns look consistent. This one is visible in the default theme too.

The sidebar showed hovered and selected rows at the same full sidebar-accent in light mode, so hovering an item looked identical to the active one. The dark-mode treatment already softened non-active hover to a half-opacity accent; that now applies in light mode too, and the reveal-overlay token is flattened to the matching opaque colour so the shortcut overlay still lines up with the row.

The password strength meter drew its segment gaps with ring-background, which only matches the surface when the page background equals the card colour. The meter always sits inside a card, so it now uses ring-card in both modes. No visual change with the default theme where background and card are both white, but it fixes forks with a tinted page background.

No visual change to the default theme except the select padding and the softer light-mode sidebar hover.

Regression guard: not warranted, these are CSS-only class tweaks (hover opacity, dropdown padding, ring surface) with no logic or data contract a test could pin without brittle pixel snapshots.


stickerdaniel:
* **#568** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/568?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/568?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #569, opened 2026-06-15
style(sidebar): raise non-active hover to 65% accent

stickerdaniel:
Follow-up to #568. The 50% non-active row hover read too faintly on near-white sidebars, so hover was hard to distinguish from the resting row. Raise it to 65% of the accent so the hover is clearly visible while still sitting below the full-accent active row, and bump the reveal-overlay token to the matching 65% mix so the shortcut overlay stays aligned with the row.

Regression guard: not warranted, single opacity-value tweak with no logic or data contract a test could pin without a brittle pixel snapshot.


stickerdaniel:
* **#569** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/569?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/569?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #570, opened 2026-06-15
test(i18n): guard against orphan keys and prune dead ones

stickerdaniel:
en.json carried 138 leaf keys (16% of 869) that no code references. The existing `scripts/i18n-used-keys.test.ts` only checks the forward direction, that every referenced key exists, so dead keys accumulated silently and every locale file had to keep mirroring them under locale parity.

This adds `scripts/i18n-orphan-keys.test.ts`, the reverse guard: every leaf key in en.json must be reachable from code. The hard part is that keys are referenced dynamically as well as literally, so a naive detector would flag live keys as dead. It resolves the dynamic forms the codebase actually uses: dotted string literals (which covers the `translationKey`/`roleKey`/`titleKey` config-property conventions, ternary branches, and lookup maps), `keyName="prefix.{x}"` interpolation and template-literal prefixes, and member-access chains like `translations.error_page.<leaf>`. Five Better Auth runtime error keys surfaced only through `getAuthErrorKey()`/`formError` are covered by a small documented allowlist. A literal-count floor makes the scan fail loudly if it breaks rather than passing vacuously by flagging everything.

With the guard in place, the 138 keys it surfaced are pruned across all four locales. They are mostly superseded namespaces (`aria.pagination_*` replaced by `table.pagination.*`, `aria.select_*` by `admin.users.select_*`) and residue from removed UI (generic `admin.settings.save/saved/saving`, dead `founder_welcome` variants, `nav.features`/`nav.solution`). Each removed key was independently confirmed to have no literal, config-value, or interpolation reference anywhere in the tree.

Regression guard: added (`scripts/i18n-orphan-keys.test.ts`), verified to go red against an injected fake orphan and to fail loudly if the scan breaks.

Verification: `bun run test:unit` (580 tests) and the three i18n guards (orphan, used-keys, parity) pass. Tolgee `i18n:cleanup` should run after merge to deprecate the removed keys in the cloud.

### DIFF
--- a/scripts/i18n-orphan-keys.test.ts
+++ b/scripts/i18n-orphan-keys.test.ts
@@ -1,0 +1,253 @@
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+import { describe, expect, it } from 'vitest';
+import en from '../src/i18n/en.json';
+
+// Defined-but-unused guard for translation keys: every leaf key in en.json must
+// be reachable from code. An orphan (a key nobody references) is dead weight
+// that every locale has to keep in sync (locale-parity.test.ts forces de/es/fr
+// to mirror en.json), and it lets a stale namespace masquerade as live copy.
+//
+// This is the mirror image of i18n-used-keys.test.ts: that test asserts every
+// referenced key EXISTS in en.json (catches typos / renamed keys); this test
+// asserts every key in en.json IS referenced (catches abandoned keys).
+//
+// The hard part is that translation keys are referenced both statically and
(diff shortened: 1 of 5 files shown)


stickerdaniel:
* **#570** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/570?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/570?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #571, opened 2026-06-15
chore(audit): consistency, docs, and svelte best-practice fixes

stickerdaniel:
A multi-dimension scan of the codebase surfaced 26 verified findings: documentation that pointed at routes, headers, and setup steps that no longer match the code, plus small consistency gaps in i18n, form validation, locale formatting, and Convex query parallelism. None were correctness or security bugs, but they accumulate as fork confusion and minor UX and accessibility rough edges.

This branch applies all 26 as seven topical commits. The documentation pass corrects the removed /emails preview route, the Vary/CDN caching wording, the CSP example, the stale proxy-worker paths, and the add-a-language steps, and registers the missing env vars and btca resources. The code pass localizes hardcoded labels (AI chat tool panel, marketing logo alt text, the admin ban-reason label) and gives the attachment-preview fallback an accessible name; parallelizes independent Convex reads in the admin dashboard and support threads; shares a single date-fns locale helper; makes the AI chat suggestions re-localize on a language switch and adds the missing effect cleanup; aligns the settings forms with the auth forms (novalidate plus Valibot name validation); and drops the legacy authParamsSchema alias.

Regression guard: not warranted, these are one-off consistency and doc-drift cleanups rather than a recurring bug class, and the existing locale-parity, prerender-sync, and orphan-key tests already guard the i18n surface.

In the branch, bun run check:convex, static-checks across all changed files, bun run test:unit (580 tests), and a full bun run build all pass.

PR #573, opened 2026-06-15
fix(btca): correct zxcvbn-ts core searchPath

stickerdaniel:
The btca resource for `@zxcvbn-ts/core` (added in #571) set `searchPath` to `packages/core`, but the zxcvbn-ts monorepo keeps the core library under `packages/libraries/main` (language packs live under `packages/languages/*`). The wrong path scoped btca searches to a directory that does not exist, so any zxcvbn lookup returned nothing.

Point `searchPath` at `packages/libraries/main` and note the language-pack location.

Regression guard: not warranted, one-off config path correction rather than a recurring class, and no test covers btca.config.jsonc resource paths.

Verified the corrected path resolves in the cloned sandbox (`packages/libraries/main/package.json` is `@zxcvbn-ts/core`); pre-commit static-checks pass.

### DIFF
--- a/btca.config.jsonc
+++ b/btca.config.jsonc
@@ -455,8 +455,8 @@
 			"name": "zxcvbnTs",
 			"url": "https://github.com/zxcvbn-ts/zxcvbn",
 			"branch": "master",
-			"searchPath": "packages/core",
-			"specialNotes": "@zxcvbn-ts/core package (monorepo). Password strength estimation. Powers the password meter. Core scoring lives in packages/core."
+			"searchPath": "packages/libraries/main",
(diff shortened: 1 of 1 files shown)


stickerdaniel:
* **#573** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/573?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/573?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #574, opened 2026-06-15
fix(admin): localize users Created column date

### LINKED ISSUE #572: admin users "Created" column ignores the active UI language
The "Created" date column in the admin users table formats with no locale, so it falls back to the browser locale instead of the active UI language. `src/routes/[[lang]]/admin/users/columns.ts:166` does `new Date(createdAt).toLocaleDateString()` with no locale argument, while the other localized routes pass `page.data.lang ?? 'en'` (for example `src/routes/[[lang]]/app/community-chat/+page.svelte:123`). On a non-English UI the created date can render in the wrong language.

It was left out of #571 because `columns` is a static `const` (`columns.ts:23`) consumed in `src/routes/[[lang]]/admin/users/+page.svelte` at lines 245, 562, 609, and 619, so threading the language needs a small factory refactor rather than a one-liner.

proposed fix: turn `columns` into a `createColumns(lang)` factory and derive it in the page.

```diff
-export const columns: Array<ColumnDef<AdminUserData>> = [
+export function createColumns(lang: string): Array<ColumnDef<AdminUserData>> {
+	return [
 		// ...
-		const formatted = createdAt ? new Date(createdAt).toLocaleDateString() : '-';
+		const formatted = createdAt ? new Date(createdAt).toLocaleDateString(lang || 'en') : '-';
 		// ...
-];
+	];
+}
```

```diff
-import { columns } from './columns.js';
+import { createColumns } from './columns.js';
+const columns = $derived(createColumns(page.data.lang ?? 'en'));
```

`columns.length` usages stay unchanged since the derived value is still an array.

stickerdaniel:
Closes #572

The admin users table "Created" column formatted dates with `new Date(createdAt).toLocaleDateString()` and no locale argument, so it fell back to the browser locale instead of the active UI language. On a non-English UI the created date could render in the wrong language.

`columns` was a static `const` consumed in `+page.svelte` both for the table and for the `columns.length` colspans, so threading the language needed a small factory refactor rather than a one-liner. This turns `columns` into a `createColumns(lang)` factory, derives `const columns = $derived(createColumns(data.lang ?? 'en'))` in the page, and passes it into `createSvelteTable` through a `get columns()` getter so the dates re-render on language switch, matching the existing `get data()` reactive-options pattern. The three `columns.length` colspan usages stay unchanged. This follows the same `data.lang ?? 'en'` approach already used by `community-chat/+page.svelte`.

Regression guard: not warranted, this was the last UI-facing toLocale* call missing a locale (the other localized routes already pass the active language) and a banned-pattern guard would be false-positive-prone against correct variable-locale and server-side en-US calls.

`bun scripts/static-checks.ts` on both changed files passes.

### DIFF
--- a/src/routes/[[lang]]/admin/users/+page.svelte
+++ b/src/routes/[[lang]]/admin/users/+page.svelte
@@ -30,12 +30,14 @@
 	import ConvexCursorTableShell from '$lib/components/tables/convex-cursor-table-shell.svelte';
 	import { createConvexCursorTable } from '$lib/tables/convex/create-convex-cursor-table.svelte';
 	import type { CursorListResult } from '$lib/tables/convex/contract';
-	import { columns } from './columns.js';
+	import { createColumns } from './columns.js';
 	import DataTableFilters from './data-table-filters.svelte';
 	import type { ActionEvent } from './data-table-actions.svelte';
 	import { browser } from '$app/environment';
 
-	let { data: _data }: { data: PageData } = $props();
+	let { data }: { data: PageData } = $props();
+
+	const columns = $derived(createColumns(data.lang ?? 'en'));
 
 	const client = useConvexClient();
 
@@ -242,7 +244,9 @@
 		get data() {
 			return usersTable.rows;
 		},
-		columns,
+		get columns() {
+			return columns;
+		},
 		state: {
 			get pagination() {
 				return { pageIndex, pageSize };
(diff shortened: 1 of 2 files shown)


stickerdaniel:
* **#574** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/574?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/574?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #547, opened 2026-06-11
fix(support): Read user identity from session cookies

stickerdaniel:
Closes #452

`getUserId()` in `customer-support.svelte` read `page.data.viewer` to identify the signed-in user. The widget is mounted on the prerendered marketing home page, where `page.data.viewer` is baked to `null` at build time and never recovers, while `auth.isAuthenticated` does recover via cookies. A signed-in visitor therefore passed the auth gate, got no viewer id, and was handed a fresh anonymous id, keying their support threads to the wrong identity. `feedback-widget.svelte` prefilled the notification email from the same frozen source.

Both components now subscribe to `authClient.useSession()`, which recovers via session cookies on prerendered pages, matching the pattern already used by `marketing-header.svelte` and `support-ticket-migration-bootstrap.svelte` and the prerendering constraints in `AGENTS.md`. `getUserId()` reads the session user id, and the `defaultEmail` prefill reads the session email. The userId effect additionally waits for a signed-in user's session to resolve before falling back, so the brief window where the session query is still pending no longer mints and persists a spurious anonymous id in localStorage. The now unused `page` import in `feedback-widget.svelte` is removed.

`bun scripts/static-checks.ts` passes on both changed files (0 errors, only pre-existing warnings in unrelated files).

PR #575, opened 2026-06-15
fix(auth): break app:auth invalidate loop on prerender

stickerdaniel:
For logged-in users the support widget on the prerendered marketing home intermittently stays blank: no thread list and no "Hi 👋" greeting, regardless of whether the user has chats. Anonymous visitors are unaffected. It loads on some page loads and stays blank on others.

The prerendered root layout bakes `authState.isAuthenticated=false` at build time. After the client recovers a session from cookies, `AppAuthProvider`'s effect saw the client-vs-server auth divergence and called `invalidate('app:auth')` to refresh `viewer`/`autumn` data. On a prerendered page the re-run returns the frozen build-time data, so the divergence never cleared and the effect re-fired on every resulting `page.data` change. Each `invalidate` re-ran the convex-better-auth provider, re-calling `client.setAuth()` and re-pausing the Convex WebSocket, so the support widget's auth-gated `listThreads` query (logged-in users carry no `anonymousUserId` since #547) stayed pending and `threads-overview` rendered its empty loading body. Whether a given load lands the query in an open or a paused window is pure timing, which is why it is intermittent. In dev the page is served dynamically, so the re-run sees the real cookie state and the divergence clears, which is why this only surfaces in production.

Guard the effect to invalidate at most once per client-auth transition, so a frozen server state can no longer drive a re-invalidate loop. The legitimate one-shot refresh on real auth transitions (sign-in and sign-out) is preserved.

Regression guard: not warranted, singular logic error at the only `invalidate('app:auth')` call site (no recurring class to guard).

`bun scripts/static-checks.ts` passes. Reproduces only on a production prerender; the preview deploy lets the logged-in marketing-home support widget be confirmed.

### DIFF
--- a/src/lib/components/app/app-auth-provider.svelte
+++ b/src/lib/components/app/app-auth-provider.svelte
@@ -16,17 +16,31 @@
 
 	const auth = useAuth();
 
+	// Track the auth value we last invalidated for, so the effect fires once per
+	// divergence instead of looping. Plain let (not $state): the effect must read
+	// it without depending on it, and write it without re-triggering itself.
+	let lastInvalidatedAuth: boolean | undefined;
+
 	// Sync server layout data when client auth state diverges.
 	// Prerendered pages bake authState.isAuthenticated: false at build time.
 	// When the client recovers a session from cookies, the server's root layout
 	// data (viewer, autumnState) remains stale. This invalidation triggers a
 	// re-run of the root +layout.server.ts with fresh cookies, so navigating
 	// to /app has the correct viewer data instead of null.
+	//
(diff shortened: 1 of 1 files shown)


stickerdaniel:
* **#575** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/575?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/575?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #578, opened 2026-06-15
fix(e2e): cache-bust public navigations and harden a11y audit

### LINKED ISSUE #576: public-a11y E2E audits a stale edge-cached build (missing cache-buster)
`e2e/public-a11y.spec.ts:25` navigates to clean URLs (`/en`, `/en/pricing`, `/en/about`, ...) with `await page.goto(path)`. CF's Cache API ignores the `Vary` header, so the post-deploy production `E2E Tests` run can be served a stale edge-cached build for those URLs, or a markdown body poisoned by `e2e/public-agent-surface.spec.ts` (which requests `/en/pricing` with `Accept: text/markdown`). axe then audits stale DOM and reports violations against elements that no longer exist in the current build.

Observed in practice: a post-deploy run reported color-contrast failures on a pricing tier button and footer links that had already been changed in the deployed build, because the cached HTML predated the fix. The PR preview run passed on the same code, since its edge cache was cold.

`e2e/upgrade-checkout-failure.spec.ts:78` already avoids this by navigating to `/en/pricing?cb=${Date.now()}` and documents the quirk at `:72-77`, but `public-a11y.spec.ts` and `public-agent-surface.spec.ts` do not.

proposed fix: append a unique `?cb=${Date.now()}` to each navigation in `public-a11y.spec.ts:25`, matching the checkout spec. The query param does not change the rendered DOM, so axe results are unchanged apart from bypassing the cache. Consider the same for `public-agent-surface.spec.ts`.

stickerdaniel:
Closes #576

Cloudflare's Cache API ignores the `Vary` header, so the post-deploy production E2E run can be served a stale edge-cached HTML build for the public marketing URLs, or a markdown body that `public-agent-surface.spec.ts` cached for the same URL with `Accept: text/markdown`. axe then audits stale or wrong DOM and reports violations against elements that are no longer in the current build, while the PR preview run passes on the same code because its edge cache is cold. I confirmed the mechanism against the live production deployment: a plain GET of `/en/pricing` returns `cf-cache-status: HIT` under `s-maxage=3600`, while the same URL with a unique `cb` query returns a fresh origin fetch, and the `cb` param is DOM-neutral (the response body is byte-identical and the canonical URL never echoes the query string).

This appends a unique cache-buster query param to every public navigation in `public-a11y.spec.ts` and `public-agent-surface.spec.ts`, including the markdown loop that was the poisoner, matching the existing workaround in `upgrade-checkout-failure.spec.ts`. The exact-URL assertions in `public-agent-surface.spec.ts` are loosened to compare the pathname, since a query string is now present.

While verifying, I found a second, pre-existing flake in `public-a11y.spec.ts` that is independent of caching: it called `makeAxeBuilder().analyze()` immediately after `page.goto()`, and axe's per-frame `page.evaluate()` passes intermittently raced the page's still-settling client bootstrap, throwing `Execution context was destroyed, most likely because of a navigation` on different pages each run (reproducible on plain `main` with no cache-buster). The spec now waits for `networkidle` and a rendered `#main-content` before auditing, and retries `analyze()` only on that specific error, re-settling between attempts. The retry cannot mask a real violation, which is returned in `results` rather than thrown, so the strict empty-violations assertion is unchanged.

Regression guard: not warranted, both changes are E2E test-hygiene with no app-code change; a static lint for "public navigations must cache-bust" is defeated by variable-driven paths and would false-positive across cacheable marketing versus non-cacheable auth routes.

Verification: `bun scripts/static-checks.ts` on both specs passes; a stress run of `public-a11y.spec.ts` at `--repeat-each=3` (42 executions across both themes) passes 42/42 with zero `Execution context was destroyed`, versus 1 to 6 failures per plain run before the fix.

### DIFF
--- a/e2e/public-a11y.spec.ts
+++ b/e2e/public-a11y.spec.ts
@@ -28,8 +28,39 @@ for (const theme of ['light', 'dark'] as const) {
 				// A unique cb per run forces a fresh origin fetch so axe audits the
 				// current build.
 				await page.goto(`${path}?cb=${Date.now()}`);
-				const results = await makeAxeBuilder().analyze();
-				expect(results.violations).toEqual([]);
+
+				// Settle before axe injects. page.goto resolves on the DOM `load`
+				// event, but the client keeps working after that (SvelteKit
+				// hydration, Convex/auth client bootstrap over the WebSocket).
+				// networkidle waits past the HTTP part of hydration; #main-content
+				// is server-rendered on every audited page (both layout groups), so
+				// the auto-retrying assertion confirms the page actually rendered.
+				await page.waitForLoadState('networkidle');
+				await expect(page.locator('#main-content')).toBeVisible();
+
+				// axe's analyze() runs many page.evaluate() passes; if the page's
+				// execution context is torn down between two of them it throws
+				// "Execution context was destroyed, most likely because of a
+				// navigation". networkidle ignores WebSocket frames, so a late
+				// socket-driven update can still trigger this; retry only that
+				// specific error, re-settling between attempts. A real a11y
+				// violation is returned in results, never thrown, so this cannot
+				// mask one: violations are still asserted empty below.
+				let results: Awaited<ReturnType<ReturnType<typeof makeAxeBuilder>['analyze']>> | undefined;
+				for (let attempt = 0; attempt < 3; attempt++) {
+					try {
+						results = await makeAxeBuilder().analyze();
+						break;
+					} catch (error) {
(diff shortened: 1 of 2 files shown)


stickerdaniel:
* **#578** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/578?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/578?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #580, opened 2026-06-16
feat(env): adopt Convex typed env declarations

### LINKED ISSUE #577: Simplify env handler with new convex functionality
https://youtu.be/fefkbux1xQo

stickerdaniel:
Closes #577

Convex 1.39 added typesafe environment variables: you declare them in `convex.config.ts`, the backend validates their presence and validators at push time, and a typed `env` object is generated into `_generated/server`. This repo predates that and instead carried a hand-rolled access layer where `env.ts` derived a `RequiredKeys` mapped type from the varlock-generated `convex-env.d.ts`, so the required/optional split lived in a second place and nothing guarded the deployment itself.

This bumps `convex` 1.37.0 to 1.40.0 (staying below 1.41 to avoid the `@convex-dev/resend` peer hold) and declares the full env var set in the `defineApp` `env` block. Env values are always strings, so only string-like validators are used. `env.ts` now sources its required/optional split from the generated `env` and reads through it, dropping the dependency on the varlock type. The module-load placeholder fallback stays, because `env` is a plain alias of `process.env`, so a top-level read is still `undefined` during Convex's static-analysis pass, which the Better Auth adapter and the Autumn client rely on.

varlock (`.env-convex.schema`) stays canonical for log redaction, leak scanning, the `@type=url`/`@type=boolean` coercion, and the SvelteKit runtime side, none of which the Convex feature covers. The var set now lives in two declarations, so a parity test asserts the schema and the `env` block agree on names and optionality.

One behavior change to watch: the `env` block makes every declared required var mandatory at push for every deployment, so a preview or production deploy that is missing one will now fail loudly at deploy time instead of at runtime. That is the point of the feature, but it means any required var must be set as a preview/prod default.

Regression guard: added `scripts/env-schema-parity.test.ts`, which asserts `.env-convex.schema` and the `convex.config.ts` env block declare the same vars with matching required/optional status.
`bun run check:convex`, `bun scripts/static-checks.ts`, and `bun run test:unit` (584 tests) pass.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -79,7 +79,7 @@
         "bits-ui": "2.18.0",
         "bun-tasks": "0.1.2",
         "clsx": "2.1.1",
-        "convex": "1.37.0",
+        "convex": "1.40.0",
         "convex-vite-plugin": "0.4.0",
         "eslint": "10.4.1",
         "eslint-config-prettier": "10.1.8",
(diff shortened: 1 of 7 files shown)


stickerdaniel:
* **#580** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/580?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/580?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #584, opened 2026-06-16
feat(emails): load Outfit webfont in email header

### LINKED ISSUE #583: Emails reference the Outfit brand font but never load it
Every email template puts `font-sans` on `<Body>`, and the email build inlines that as `font-family:'Outfit', ui-sans-serif, system-ui, sans-serif, ...` on the html element and the body cells. You can see this in any generated template, for example `src/lib/convex/emails/_generated/verification.ts`.

The problem is that the emails never declare an `@font-face` for Outfit. The website self-hosts it in `src/routes/layout.css` (the three `outfit-v15-latin-*.woff2` files in `static/fonts/`), but the email `<head>` only contains meta tags. So Outfit is named in the font stack but the file is never loaded, and every recipient falls back to their system sans (Arial/Helvetica). The brand sans shows up nowhere in the actual mail, and this affects every fork.

Suggested fix: a shared `EmailHead` component that wraps `<Head>` and injects an `@font-face` for Outfit, reusing the woff2 the site already self-hosts (via the same `__BASEURL__` rewrite the logo uses), swapped in for the bare `<Head />` in the six templates. This is progressive enhancement: Apple Mail and iOS Mail load the brand font, while Gmail and Outlook keep the existing system-sans fallback that is already in the stack. The rule has to be injected as a string with `{@html}` since a Svelte `<style>` block gets scoped and never reaches the head.

stickerdaniel:
Closes #583

The email templates set `font-sans` on the body, which the build inlines as `font-family:'Outfit', ...`, but the email `<head>` never declared an `@font-face` for Outfit. The website self-hosts the font, but emails did not, so the brand sans never loaded and every client fell back to its system sans.

This adds a shared `EmailHead` component that wraps `@better-svelte-email`'s `<Head>` and injects an `@font-face` for Outfit, reusing the three `outfit-v15-latin-*.woff2` files the site already self-hosts, served through the same `__BASEURL__` rewrite the logo uses. The six templates now use `<EmailHead />` instead of a bare `<Head />`. It is progressive enhancement: Apple Mail and iOS Mail render the brand font, while Gmail, Outlook, and the rest keep the system-sans fallback already in the stack. The rule is injected as a string with `{@html}` because a Svelte `<style>` block compiles to scoped component CSS and never reaches the head, and `mso-generic-font-family` keeps Outlook Windows on a sans fallback instead of dropping to Times New Roman.

`bun scripts/static-checks.ts` passes and the email snapshot tests pass (snapshots refreshed to include the new font declaration).


stickerdaniel:
* **#584** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/584?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/584?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #585, opened 2026-06-19
feat(skills): add upstream-sync skill for pulling template changes

stickerdaniel:
Forks of this template are created by content-copy ("Use this template"), so they share no git ancestor with upstream. That makes `git merge`, `git rebase`, and `gh repo sync` useless for pulling later template changes into a fork, and there was no tooling to do it: every fork had to re-derive the whole procedure by hand.

This adds an `upstream-sync` skill that encodes the procedure. Two read-only discovery scripts do the deterministic part: `find-fork-point.ts` recovers the synthetic merge-base by tree-SHA identity (the bootstrap commit copied an upstream tree verbatim, so its tree SHA equals one upstream commit's), and `list-upstream-changes.ts` lists every upstream commit since the last sync, oldest-first, each priority-tagged (security, fix, feat, chore) and annotated with the fork-divergence categories it touches. The `SKILL.md` body carries the judgment layer: review every commit and integrate, skip as already-present, or exclude with a recorded reason; map cross-commit dependencies; adapt anything touching the fork's branding, theme, env/deploy, i18n, or own features rather than blind-applying; resolve i18n via a JSON-aware 3-way deep merge; validate against whole-project CI rather than a file-scoped check; and ship one consolidated PR rather than a stack. Four `reference/` files hold the deep detail and load only when needed.

A committed `.upstream-sync.json` marker (written by the fork after a successful sync) makes each run incremental and records deliberate exclusions. There is a `bun run upstream:sync` alias, and an AGENTS.md pointer next to the existing template-bug section so the fix-down / report-up linkage survives a rebrand. The skill follows the repo's first-party layout (`skills/upstream-sync/` with a `.claude/skills/` symlink) and is generic: it detects each fork's divergences at runtime instead of hardcoding any.

The two scripts were validated end-to-end against a real downstream fork: `find-fork-point.ts` resolves the correct fork point by exact tree-SHA match, and `list-upstream-changes.ts` classifies the full commit range. `bun run static-checks` on both scripts and `prettier --check` on all new files pass.

### DIFF
--- a/.claude/skills/upstream-sync
+++ b/.claude/skills/upstream-sync
@@ -1,0 +1,1 @@
+../../skills/upstream-sync
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -14,6 +14,12 @@ If a project forked from this repo discovers a bug or weird behavior that origin
 2. If none exists, fetch the issue template from the upstream repo first (`gh api repos/stickerdaniel/saas-starter/contents/.github/ISSUE_TEMPLATE` and read the relevant template) so the new issue follows the expected format.
 3. File a new issue describing the template bug/unexpected behavior using that template.
 
+### Pulling Upstream Template Changes
+
+<!-- DO NOT rename or remove this section when rebranding a fork. Fixes flow down from the template here; bug reports flow up via the section above. -->
+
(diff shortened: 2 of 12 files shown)


stickerdaniel:
* **#585** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/585?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/585?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #593, opened 2026-06-23
fix(e2e): stop reloading the auth wait on cold previews

### LINKED ISSUE #592: E2E waitForAuthenticated reloads on a too-small budget and cascade-fails on cold previews
`e2e/utils/auth.ts` waits for the authenticated app shell with `timeout=30000, maxRetries=3`, so each attempt only gets about 10s. It waits for `#user-menu-trigger` before `html[data-hydrated]`, and on timeout it calls `page.reload()` and retries.

On a cold Cloudflare preview this is self-sabotaging. The preview boots a fresh worker and a freshly provisioned Convex deployment, and signin navigates with a full page load, so `/app` starts a cold client and WebSocket. A `page.reload()` throws that warm boot away and re-pays the entire cold start from zero. With only about 10s per attempt and three reloads, the auth setup project times out before the preview finishes hydrating, and every dependent spec cascade-fails.

Fix: drop the retry loop and the reload. Use a single attempt on a generous budget (about 60s), wait for `html[data-hydrated]` first (it is set unconditionally in the layout's onMount, independent of the WebSocket), then `#user-menu-trigger` on a short secondary timeout. Pin Playwright `workers` to 1 everywhere (the suite shares one preview backend and admin specs mutate shared rows), and on CI raise the per-test timeout to 90s and the expect timeout to 15s so a slow-but-functional preview does not cascade into auth-setup timeouts.

stickerdaniel:
Closes #592

`waitForAuthenticated` waited for the authenticated shell with a 30s budget split across three attempts (about 10s each), checked `#user-menu-trigger` before `html[data-hydrated]`, and called `page.reload()` on timeout. On a cold Cloudflare preview that boots a fresh worker and a freshly provisioned Convex deployment, signin navigates with a full page load so `/app` starts a cold client and WebSocket. A reload throws that warm boot away and re-pays the entire cold start from zero, so with only about 10s per attempt the auth setup project timed out before the preview hydrated and every dependent spec cascade-failed.

This drops the retry loop and the reload. The helper now runs a single attempt on a generous 60s budget, waits for `html[data-hydrated]` first (set unconditionally in the layout's onMount, so it is independent of the Convex WebSocket) and then `#user-menu-trigger` on a short secondary timeout. The Playwright config pins `workers` to 1 everywhere, since the suite shares one preview backend and admin specs mutate shared rows, and raises the CI per-test timeout to 90s and the expect timeout to 15s so a slow-but-functional preview has headroom instead of cascading into auth-setup timeouts.

Regression guard: not warranted, one-off test-harness timing fix with no other call sites.

`bun run test:unit` passes.

### DIFF
--- a/e2e/utils/auth.ts
+++ b/e2e/utils/auth.ts
@@ -1,40 +1,27 @@
 import { expect, type Page } from '@playwright/test';
 
 /**
- * Wait for the app to be fully loaded, hydrated, and authenticated.
- * Handles transient app-load failures in CI by retrying page load.
+ * Wait for an authenticated app page to be loaded, hydrated, and interactive.
  *
- * 1. Wait for URL to match the app route
- * 2. Wait for the authenticated UI shell to be visible
- * 3. Wait for Svelte hydration (data-hydrated attribute on <html>)
- * 4. If the shell does not appear, reload and retry (up to maxRetries)
+ * Two signals, in order, on a single generous budget, and deliberately NO
+ * page.reload() on the way:
+ *   1. html[data-hydrated] is set unconditionally in the layout's onMount, so it is
+ *      independent of the Convex WebSocket. The definitive "Svelte hydrated, event
+ *      handlers attached" signal.
+ *   2. #user-menu-trigger is the authenticated shell, SSR-rendered from the JWT
+ *      viewer, so it paints without waiting on the live session.
+ *
+ * Why no reload and a large budget: a cold CF preview boots a fresh worker AND a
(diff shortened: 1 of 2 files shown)


stickerdaniel:
* **#593** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/593?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/593?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Issue #458, opened 2026-06-04
Auth-dependent SSR loads skip the try-catch resilience pattern


stickerdaniel:
The root layout wraps its SSR Convex queries in `.catch` fallbacks so a transient backend failure degrades instead of failing the request. Four loads skip that wrap, so the same hiccup 500s pages that would hydrate fine via their `useQuery` initialData fallback:

- [ ] src/routes/[[lang]]/(auth)/signin/+page.server.ts:7
- [ ] src/routes/[[lang]]/(auth)/signup/+page.server.ts:7
- [ ] src/routes/[[lang]]/app/ai-chat/+page.server.ts:8
- [ ] src/routes/[[lang]]/app/community-chat/+page.server.ts:8

proposed fix: `.catch(() => null)` (or `{}` for the oauth provider map) per the +layout.server.ts pattern, letting the client subscription recover after hydration.

PR #595, opened 2026-06-23
fix(ssr): guard server Convex client construction in SSR loads

### LINKED ISSUE #594: Auth-group SSR loads build the server Convex client outside the try-catch, so construction failures 500 the page
Issue #458 wrapped the SSR Convex *queries* in `.catch` so a transient backend hiccup degrades instead of failing the request. But the client itself is still built outside that guard: `signin/+page.server.ts` and `signup/+page.server.ts` call `createServerConvexHttpClient({})` on a bare line, and the root `+layout.server.ts` authenticated block builds `createServerConvexHttpClient({ token })` and sets up the Autumn handler before any `.catch`.

`createServerConvexHttpClient` throws if the Convex URL is unresolved (a missing `CONVEX_INTERNAL_URL`/`PUBLIC_CONVEX_URL`, which happens on a freshly created preview before env propagates). That throw is outside the per-query `.catch`, so it 500s every auth SSR page and every authenticated page, even though the client `useQuery` subscription would recover the real data after hydration.

Fix: wrap the whole load (construction + query) in try-catch. The two auth loads degrade to `{ oauthProviders: { google: false, github: false } }`; the layout's authenticated block degrades to the JWT viewer + null customer.

stickerdaniel:
Closes #594

Issue #458 wrapped the SSR Convex queries in a `.catch` so a transient backend hiccup degrades instead of 500ing the request, but the client construction stayed outside that guard. `createServerConvexHttpClient` throws synchronously when the Convex URL is unresolved, which happens on a freshly created preview before env propagates, and because that throw is outside the per-query `.catch` it 500s the page even though the client `useQuery` subscription would recover the real data after hydration. Five SSR loads had this shape: the signin and signup loads and the root `+layout.server.ts` authenticated block, plus the ai-chat and community-chat page loads.

This wraps construction and query together in try-catch across all five. The signin and signup loads degrade to `{ oauthProviders: { google: false, github: false } }`, the layout's authenticated block degrades to the JWT viewer with a null customer, ai-chat degrades to a null viewer, and community-chat to a null viewer with empty messages. The live subscriptions then heal the state once the client connects after hydration.

A new authenticated load could reintroduce the bare-construction pattern, which is invisible in local dev where the Convex URL always resolves and only surfaces on cold previews. A local ESLint rule now requires the construction to sit inside a try in `+page.server.ts`/`+layout.server.ts` loads.

Regression guard: added eslint/rules/require-guarded-server-convex-client.js, which flags `createServerConvexHttpClient` constructed outside a try in server loads. It is unit-tested and verified to flag the pre-fix ai-chat and community-chat loads.

`bun scripts/static-checks.ts` on the changed files passes and the new rule's unit tests pass.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -19,6 +19,7 @@ import noBareTestSkipRule from './eslint/rules/no-bare-test-skip.js';
 import noModuleStateSingletonRule from './eslint/rules/no-module-state-singleton.js';
 import requireMotionGuardTransitionRule from './eslint/rules/require-motion-guard-transition.js';
 import requireFieldErrorAssociationRule from './eslint/rules/require-field-error-association.js';
+import requireGuardedServerConvexClientRule from './eslint/rules/require-guarded-server-convex-client.js';
 
 const gitignorePath = path.resolve(import.meta.dirname, '.gitignore');
 const localPlugin = {
@@ -33,7 +34,8 @@ const localPlugin = {
 		'no-bare-test-skip': noBareTestSkipRule,
 		'no-module-state-singleton': noModuleStateSingletonRule,
 		'require-motion-guard-transition': requireMotionGuardTransitionRule,
-		'require-field-error-association': requireFieldErrorAssociationRule
+		'require-field-error-association': requireFieldErrorAssociationRule,
+		'require-guarded-server-convex-client': requireGuardedServerConvexClientRule
 	}
 };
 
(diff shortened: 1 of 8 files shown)


stickerdaniel:
* **#595** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/595?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/595?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #596, opened 2026-06-23
fix(env): mark Sentry org/project slugs public, not sensitive

### LINKED ISSUE #591: Sentry org/project slugs classified sensitive, false-positive leak breaks preview builds
`.env.schema` marks `SENTRY_ORG` and `SENTRY_PROJECT` as `@optional` but not `@public`. With the schema-level `@defaultSensitive=inferFromPrefix(PUBLIC_)` plus `@preventLeaks=true`, any var without a `PUBLIC_` prefix is treated as a secret, so both Sentry slugs are classified sensitive.

They are not secrets: the org and project slugs are public and are embedded in the client via the DSN. The harm shows up on preview builds. When a slug value also appears legitimately in the build output, varlock's leak scanner false-positives and fails the build. In a fork where `SENTRY_PROJECT` equals the Cloudflare Worker name, the preview URL `<alias>-<worker>.workers.dev` contains the slug, so every preview build fails the leak scan, while production (custom domain, no match) builds fine. The result is that all PR preview builds and their E2E runs break.

Fix: mark both `@public @optional`. Only `SENTRY_AUTH_TOKEN` is a real secret. This also regenerates `src/env.d.ts` to drop the sensitive flag on the two slugs.

stickerdaniel:
Closes #591

`.env.schema` marked `SENTRY_ORG` and `SENTRY_PROJECT` as `@optional` but not `@public`. With the schema-level `@defaultSensitive=inferFromPrefix(PUBLIC_)`, any var without a `PUBLIC_` prefix is treated as a secret, so both Sentry slugs were classified sensitive even though they are public values embedded in the client via the DSN. The harm shows up on preview builds: when a slug value also appears in the build output (for example `SENTRY_PROJECT` equal to the deployed worker name, which shows up in the preview URL), varlock's leak scanner false-positives and fails the build, breaking every PR preview and its E2E run, while production builds on a custom domain pass.

This marks both `@public @optional`; only `SENTRY_AUTH_TOKEN` remains a real secret. Running varlock typegen regenerates `src/env.d.ts` to drop the sensitive flag on the two slugs and add them to the public schema.

Regression guard: covered by varlock's own leak scan, which now treats the slugs correctly; a separate guard would duplicate the schema-to-typegen contract.

`bun install` regenerates `src/env.d.ts` cleanly and the pre-commit checks pass.

### DIFF
--- a/.env.schema
+++ b/.env.schema
@@ -93,12 +93,15 @@ PUBLIC_SENTRY_DSN=
 # @sensitive @optional
 SENTRY_AUTH_TOKEN=
 
-# Sentry organization slug (build-time only)
-# @optional
+# Sentry organization slug (build-time only). Public slug, not a secret (it is
+# part of the public DSN). Marked @public so varlock does not flag it as a
+# leaked secret: when SENTRY_PROJECT equals the deployed worker name it
+# legitimately appears in preview-build URLs and tripped the leak scanner.
+# @public @optional
 SENTRY_ORG=
 
-# Sentry project slug (build-time only)
-# @optional
+# Sentry project slug (build-time only). Public slug, not a secret. See SENTRY_ORG.
+# @public @optional
 SENTRY_PROJECT=
 
 # ====================================
--- a/src/env.d.ts
+++ b/src/env.d.ts
@@ -115,15 +115,18 @@ export type CoercedEnvSchema = {
   SENTRY_AUTH_TOKEN?: string;
   
   /**
-   * **SENTRY_ORG** 🔐 _sensitive_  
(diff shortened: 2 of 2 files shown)


stickerdaniel:
* **#596** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/596?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/596?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #598, opened 2026-06-23
fix(build): bake Sentry only into production builds

### LINKED ISSUE #597: Sentry env vars bake into preview builds because CF Workers Builds shares build env across triggers
`vite.config.ts` enables the Sentry plugin whenever `PUBLIC_SENTRY_DSN` is present and otherwise passes the Sentry env through unchanged. CF Workers Builds shares the same build environment variables across the production and preview triggers, so `PUBLIC_SENTRY_DSN` (and the auth token / slugs) bake into every preview and PR deploy, not just production.

Two consequences on previews: the Sentry browser SDK loads on every public page, which destabilizes the public E2E run, and runtime errors from a PR get reported into the production Sentry project, polluting it with noise from unmerged work.

Fix: blank the Sentry vars for any non-production build, detected the same way `scripts/cf-deploy.ts` does (`WORKERS_CI_BRANCH` equals `PRODUCTION_BRANCH`, default `main`). Blank rather than delete, because `PUBLIC_SENTRY_DSN` is imported from `$env/static/public` which only exports vars present at build time, so deleting it breaks the import; an empty string keeps the export, reads as falsy, and lets the SDK dead-code-eliminate.

stickerdaniel:
Closes #597

`vite.config.ts` enabled the Sentry plugin whenever `PUBLIC_SENTRY_DSN` was present and otherwise passed the Sentry env through unchanged. CF Workers Builds shares the same build environment variables across the production and preview triggers, so `PUBLIC_SENTRY_DSN` and the auth token and slugs baked into every preview and PR deploy. That loaded the Sentry browser SDK on every public page, which destabilizes the public E2E run, and reported runtime errors from unmerged PRs into the production Sentry project.

This blanks the Sentry vars for any non-production build, detecting production the same way `scripts/cf-deploy.ts` does (`WORKERS_CI_BRANCH` equal to `PRODUCTION_BRANCH`, default `main`). The vars are blanked rather than deleted because `PUBLIC_SENTRY_DSN` is imported from `$env/static/public`, which only exports vars present at build time, so deleting it would break the import with MISSING_EXPORT; an empty string keeps the export, reads as falsy, and lets the SDK dead-code-eliminate on previews.

Regression guard: not warranted, a single build-config branch with no other call sites.

`bun run build` succeeds and pre-commit checks pass.

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -188,6 +188,25 @@ function printOptionalFeatureBanner(opts: {
 export default defineConfig(async ({ mode }) => {
 	const cwd = process.cwd();
 	const loadedEnv = loadEnv(mode, cwd, '');
+
+	// Sentry is production-only. CF Workers Builds shares build env vars across the
+	// production and preview triggers, so PUBLIC_SENTRY_DSN would otherwise bake into
+	// preview/PR deploys: it loads the SDK on every public page (which destabilizes the
+	// public E2E run) and reports PR errors into the prod Sentry project. Blank the
+	// Sentry vars for any non-production build. Blank, not delete: PUBLIC_SENTRY_DSN is
+	// imported from $env/static/public, which only exports vars present at build time,
+	// so deleting it breaks that import (MISSING_EXPORT). An empty string keeps the
+	// export but reads as falsy, so the SDK is still dead-code-eliminated. Mirrors the
+	// prod/preview check in scripts/cf-deploy.ts (WORKERS_CI_BRANCH).
+	const ciBranch = process.env.WORKERS_CI_BRANCH;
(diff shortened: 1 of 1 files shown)


stickerdaniel:
* **#598** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/598?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/598?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #602, opened 2026-06-23
fix(analytics): cookieless PostHog, identify by id

### LINKED ISSUE #601: PostHog sets a cookie while the privacy policy claims no consent banner is needed
`posthog.init()` in `src/lib/analytics/posthog.ts` sets no `persistence`, so posthog-js defaults to `localStorage+cookie` and writes the `ph_<token>_posthog` cookie. But the privacy policy (`src/lib/content/privacy.ts`, Section V) states "We use only essential cookies... No consent banner is required as these cookies are strictly necessary." Under GDPR and German §25 TDDDG that is inaccurate as shipped: the analytics cookie is not strictly necessary and would require consent. Every fork inherits this.

Fix: initialize PostHog cookieless with `persistence: 'memory'`. Autocapture, pageviews, and identify keep working; only cross-reload anonymous attribution degrades. The existing privacy copy then becomes accurate and no banner is required.

Related improvement: `src/lib/components/analytics/PostHogIdentify.svelte` uses the user's email as the `distinct_id` and passes the id only as a property. The stable user id is the better default distinct_id (email can change, and the id is the join key for any server-side events); keep email and name as person properties.

stickerdaniel:
Closes #601

`posthog.init` set no `persistence`, so posthog-js defaulted to `localStorage+cookie` and wrote the `ph_<token>_posthog` cookie. The privacy policy (`src/lib/content/privacy.ts`, Section V) claims only essential cookies are used and no consent banner is required, which is inaccurate as shipped under GDPR and German §25 TDDDG, and every fork inherits it.

Setting `persistence: 'memory'` makes PostHog cookieless, so the existing privacy copy becomes accurate with no banner needed. Autocapture, pageviews, and identify keep working; only cross-reload anonymous attribution degrades. The identify call now uses the stable user id as the `distinct_id` instead of email, keeping email and name as person properties, since email can change and the id is the join key for any server-side events.

Regression guard: not warranted, a single PostHog init option at the only init site.

`bun scripts/static-checks.ts` passes on both changed files.

### DIFF
--- a/src/lib/analytics/posthog.ts
+++ b/src/lib/analytics/posthog.ts
@@ -65,7 +65,12 @@ export async function initPosthog(): Promise<PostHogClient | null> {
 			posthog.init(PUBLIC_POSTHOG_API_KEY, {
 				api_host: apiHost,
 				ui_host: 'https://eu.posthog.com',
-				person_profiles: 'identified_only'
+				person_profiles: 'identified_only',
+				// Cookieless: store nothing in cookies/localStorage so the privacy
+				// policy's "essential cookies only, no consent banner" claim holds
+				// (GDPR/§25 TDDDG). Autocapture, pageview, and identify keep working;
+				// only cross-reload anonymous attribution degrades.
+				persistence: 'memory'
 			});
 
 			client = posthog;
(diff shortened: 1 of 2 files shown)


stickerdaniel:
* **#602** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/602?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/602?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #603, opened 2026-06-24
fix(test): stub required secrets for local e2e

stickerdaniel:
A fresh `bun run dev:test` backend (a new git worktree, or after `RESET_LOCAL_BACKEND`) fails to deploy with `MissingEnvironmentVariables`. `convex.config.ts` declares `RESEND_API_KEY`, `AUTH_EMAIL`, `EMAIL_ASSET_URL`, `AUTUMN_SECRET_KEY`, and `OPENROUTER_API_KEY` as required, and Convex validates the declared env block at push time. The local embedded backend only receives env that `vite.config.ts` injects, which did not include these five, so a fresh checkout could not run e2e locally without manually setting real secrets first.

Inject inert placeholders for those five in test mode, before the `.env.convex.local` spread so real values still override. The e2e suite never exercises the underlying services, Resend skips `@e2e.example.com` recipients, Autumn fails open, and OpenRouter is unused, so the dummies satisfy Convex's push-time validation with no real secret. The placeholders live in `scripts/local-convex-env.ts` as a single source of truth. Regular `bun run dev` is unchanged, the placeholders are test-mode only, so a developer's persisted backend keys are never clobbered, and for real local email/billing/AI you still put real or sandbox keys in `.env.convex.local`. The prod deploy gate, the CI env preflight, and preview/prod env wiring are untouched. The stale `.env.convex.example` note that claimed these were optional for `bun run dev` boot is corrected.

Verified by running a public e2e spec in a fresh worktree with a clean `.env.convex.local` and no manual env: the backend deploys and the tests pass (`4 passed`). `bun run test:unit` passes, including the extended `env-schema-parity` guard.

Regression guard: added env-schema-parity coverage asserting every required `convex.config.ts` var is provided to a fresh local test backend (catches a newly added required secret that CI would miss, since CI uses real preview-default env).

### DIFF
--- a/.env.convex.example
+++ b/.env.convex.example
@@ -4,25 +4,29 @@
 # Source of truth: .env-convex.schema (varlock)
 # Platform: Convex Dashboard (via CLI: bunx convex env set KEY value)
 # Purpose: Backend configuration for Convex serverless functions
-# For local dev: put values in .env.convex.local instead
+# For local dev: put values in .env.convex.local instead.
+# For local e2e (`bun run dev:test`): the required email/billing/AI secrets below
+# are auto-stubbed with inert dummies by vite.config.ts, so a fresh test backend
+# boots with no setup. A fresh `bun run dev` backend still needs the real values
+# (Convex validates the declared-required vars at push time).
 
 # ====================================
 # Email System (Resend) - REQUIRED for cloud email flows
 # ====================================
 
-# Resend API Key (REQUIRED for cloud / OPTIONAL for `bun run dev` boot)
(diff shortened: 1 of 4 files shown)


stickerdaniel:
* **#603** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/603?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/603?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #607, opened 2026-06-25
perf(fonts): add metric fallback faces, drop orphan @fontsource

stickerdaniel:
The self-hosted Outfit had no fallback-metric matching, so the swap from the system fallback to the real face shifted layout (CLS) on first paint. Separately, @fontsource-variable/outfit sat in the dependencies but was never imported anywhere (the font is self-hosted under static/fonts), so it was dead weight.

This adds fontaine as a Vite plugin that generates a metric-matched "Outfit fallback" face from the real @font-face and appends it to the web app's own font usages, so the webfont swap shifts layout close to zero. The key constraint, kept deliberately: the fallback family is never written into the --font-* tokens, because the email renderer inlines those same tokens verbatim (src/lib/emails/renderer.ts imports layout.css as a raw string), and a web-only fallback family would otherwise leak into transactional emails where no client can resolve it. A new test in email-snapshots.test.ts fails if that ever regresses. The orphan @fontsource-variable/outfit dependency is removed.

Verification: the web body font resolves to "Outfit, 'Outfit fallback', ui-sans-serif, ..." with the fallback face generated and loaded; the generated emails contain zero "Outfit fallback"; bun run test:unit passes (635 tests, including the new boundary guard).


stickerdaniel:
* **#607** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/607?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/607?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #611, opened 2026-06-25
fix: recover from blank page after deploy

stickerdaniel:
After a deploy, a returning visitor can boot a stale app shell whose content-hashed `/_app/immutable/*` chunks were replaced. A client-side navigation then imports a chunk hash that no longer exists and the page renders blank until a manual hard refresh. The marketing HTML shell was also browser-cached for four hours (Cloudflare injects max-age=14400 because the origin sent no browser max-age), which widened the window, and the template had none of SvelteKit's version-recovery guards.

This adds the canonical SvelteKit version-recovery setup plus a reactive backstop. `svelte.config.js` sets a deterministic `kit.version.name` from the commit SHA (Workers Builds, then Vercel, then a local git rev, then `dev`) and a five minute `pollInterval`. The root layout registers the documented `beforeNavigate` guard that turns the next client navigation into a full document load once `updated.current` flips, plus a `vite:preloadError` listener that hard reloads once (sessionStorage-guarded against a loop). `handleCacheControl` and the prerendered-HTML header injected in `patch-cf-worker.ts` now send `public, max-age=0, must-revalidate, s-maxage=3600, stale-while-revalidate=86400` so the browser revalidates the shell while the edge keeps serving it.

One Cloudflare-side change is required for the header to take full effect on a deployed instance: the zone Browser Cache TTL must be set to Respect Existing Headers, otherwise the four hour default overrides the origin `max-age=0` back up to 14400. That is documented inline at both cache call sites.

Regression guard: added `scripts/version-config.test.ts` (the version config and root-layout guard contract) and `e2e/public-deploy-recovery.spec.ts` (the runtime reload backstop).

Verification: `bun run check`, `bun run lint`, `bun run test:unit` (including the new `scripts/version-config.test.ts` and the `patch-cf-worker` revalidation cases), and `bun run build` all pass. The new e2e spec passes locally against the `dev:test` stack.

### DIFF
--- a/e2e/public-deploy-recovery.spec.ts
+++ b/e2e/public-deploy-recovery.spec.ts
@@ -1,0 +1,59 @@
+import { expect, test } from '@playwright/test';
+
+// The deploy-recovery contract: after a new deploy replaces content-hashed chunks,
+// a dynamic import of a now-deleted hash makes Vite fire vite:preloadError. The root
+// layout listens for it and hard reloads once to pull the fresh shell, recovering
+// from what would otherwise be a blank page. A sessionStorage flag stops a reload
+// loop if the shell itself is still served stale.
+type SentinelWindow = Window & { __reloadSentinel?: boolean };
+
+test.describe('deploy stale-chunk recovery', () => {
+	test('a failed chunk preload triggers a single full reload', async ({ page }) => {
+		// A light prerendered page: the home hero is a heavy WebGL scene that delays
+		// hydration on a GPU-less CI runner, which would race the listener attaching.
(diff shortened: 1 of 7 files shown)


stickerdaniel:
* **#611** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/611?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/611?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #613, opened 2026-06-27
style(support): hug active filter pill to its label

stickerdaniel:
The support filter tabs (Inbox / All / Unassigned) override the shadcn Tabs.List with grid w-full grid-cols-3. That stretches each trigger to a rigid third, so the active pill fills the whole third and a short label like All floats in an oversized pill. (A plain flex w-full variant renders identically, so that is not the fix.)

This keeps the bar full width but makes each trigger size to its label (flex-none, overriding the component's flex-1) and spreads the three evenly (justify-evenly), so the active pill wraps only its label. Found while fixing the same bug in a downstream fork; verified by rendering the real Tabs component with these exact classes in isolation at mobile width.

static-checks on the file passes.

### DIFF
--- a/src/routes/[[lang]]/admin/support/thread-list.svelte
+++ b/src/routes/[[lang]]/admin/support/thread-list.svelte
@@ -159,10 +159,17 @@
 			value={filterMode}
 			onValueChange={(v) => onFilterChange(v as 'all' | 'unassigned' | 'my-inbox')}
 		>
-			<Tabs.List class="grid w-full grid-cols-3">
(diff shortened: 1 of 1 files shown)


stickerdaniel:
* **#613** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/613?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/613?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #615, opened 2026-06-28
feat(usage): record per-call LLM token cost

stickerdaniel:
Adds generic per-call LLM token and cost tracking to the template. Every fork that keeps the AI chat or support agent now gets per-call token counts and USD cost in its own database, with zero new dependencies (openrouter, ai, and the agent component are already present) and no config beyond the existing OPENROUTER_API_KEY.

One aiUsage table records a row per operation (ai_chat, ai_chat_title, support_chat) with a per-model breakdown. Native OpenRouter cost is used where the provider returns it; otherwise a vendored per-model price map computes it, and the row is stamped with whether the cost was native, computed, or unknown. The capture is wired into all five call sites: the two agents go through a per-call usage sink on the convex-agent usage handler, and the direct title generateText reads usage off its result. The table is write-only for now (no UI), ready for a fork to add an admin or billing view on the by_user_at / by_feature_at indexes.

Separately, the support agent instructions are extracted into an overridable export and a docs recipe explains how to optimize that prompt against resolved tickets, without pulling a prompt-optimization framework into the template by default.

Verification: bun run check:convex, the two new unit test files, and the static-checks gate all pass.


stickerdaniel:
* **#615** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/615?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/615?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #616, opened 2026-06-28
fix(auth): recover from blank page on auth failure

stickerdaniel:
The authenticated layout rendered nothing when the server resolved no viewer (`AuthenticatedLayout` had an `{#if user}` block with no `{:else}`), so a failed auth handshake produced a blank white page that a reload could not fix. One real-world trigger is a misconfigured device clock: a clock that is far ahead makes the browser treat the freshly minted short-lived auth cookie as already expired (RFC 6265) and drop it, so every SSR request, including reloads, sees no token. The realtime auth layer itself is clock-agnostic (Convex schedules token refresh from the token's own lifetime and validates against the server clock), so the failure surfaces below the app, at the cookie and render layers.

The authenticated layout now renders a recover state instead of nothing: a calm "connecting" view first, since the client may still recover a session from cookies right after hydration, escalating to a retry state after a short grace window. Independently, a small `/api/time` endpoint lets the client measure its offset against server time once after hydration; past a 2 minute threshold a root-level banner explains the wrong clock and tells the user to set date and time to automatic, because a reload alone will not fix it. Both surfaces use the `warning` theme token and share the skew state via context, and the banner is intentionally not a forced reload, which would loop without fixing the clock.

Regression guard: added src/lib/utils/clock-skew.test.ts (skew computation, threshold, and magnitude formatting).
Verification: static-checks pass on all changed files and the new unit test plus the locale-parity test pass.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -427,7 +427,7 @@ Decision tree for new routes:
 
 4. **Is the page an API endpoint or static file?**
    - Yes → `+server.ts` with custom Response
-   - Examples: `/api/auth/[...all]`, `/llms.txt`, `/robots.txt`, `/sitemap.xml`
+   - Examples: `/api/auth/[...all]`, `/llms.txt`, `/robots.txt`, `/sitemap.xml`, `/api/time` (runtime JSON, `prerender = false`, `cache-control: no-store` — server wall-clock for client clock-skew detection)
 
 #### Data-fetching patterns (decision tree)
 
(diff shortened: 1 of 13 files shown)


stickerdaniel:
* **#616** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/616?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/616?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #617, opened 2026-06-28
fix(brand): inline logo mark to remove first-paint flash

stickerdaniel:
The brand mark in the header was invisible on first paint while inline-SVG icons rendered instantly. The mark is a `<span>` colored by a background that a CSS `[mask-image:url(/logo.svg)]` clips to the mark shape, so it depends on fetching the external `/logo.svg` before anything shows. Until that file arrives the masked span is fully transparent, so on a cold cache the mark pops in late, while inline-SVG icons in the HTML payload never wait on a fetch.

This keeps `static/logo.svg` as the single logo source, the same file that still feeds favicon, PWA, and email generation, but inlines it into `logo.svelte` at build time with a `?raw` import instead of fetching it at runtime. The mark now ships in the HTML payload and paints with the rest of the header. The source SVG strokes in `currentColor` and the wrapper keeps the optional `--logo-color` override, so theming is unchanged. A component-scoped `:global(svg)` rule pins the inlined SVG to the wrapper box so it always matches the `size-*` utility and never falls back to its intrinsic dimensions while utility CSS catches up. A `$static` alias addresses the single source cleanly, mirroring the existing `$blocks` alias and the `?raw` pattern already used in `src/lib/emails/renderer.ts`.

Regression guard: not warranted, single-site rendering change; the `mask-image` approach lived only in `logo.svelte`.

`bun scripts/static-checks.ts` on the changed files passes.

### DIFF
--- a/src/lib/components/icons/logo.svelte
+++ b/src/lib/components/icons/logo.svelte
@@ -1,14 +1,37 @@
 <script lang="ts">
 	import type { IconProps } from '@lucide/svelte';
 	import { cn } from '$lib/utils';
+	// Inlined at build time from the single logo source (static/logo.svg), the same
+	// file that feeds favicon/PWA/email generation. Shipping the mark in the HTML
+	// payload avoids the runtime fetch of a CSS mask-image, which left the mark
+	// blank on first paint until /logo.svg loaded over the network. Trusted local
+	// build-time constant, so {@html} carries no injection risk.
+	import logoSvg from '$static/logo.svg?raw';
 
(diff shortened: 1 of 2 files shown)


stickerdaniel:
* **#617** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/617?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/617?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #618, opened 2026-06-29
fix(support): avatar fallback while images load

stickerdaniel:
The support widget welcome screen gated its avatar-stack entrance animation on a separate `Image()` preload, but the displayed shadcn `Avatar`s run their own internal load (bits-ui) and keep the `` at `display:none` until that settles. The two loads were never in sync, so the stack could animate in while a remote admin photo was still loading. With no `AvatarFallback` on those avatars, the circle rendered fully transparent, which looked broken.

This gates the entrance on each `Avatar`'s real `loadingStatus` (via `onLoadingStatusChange`) instead of the parallel preload, so "ready" now means the rendered image is actually painted. It adds an `AvatarFallback` to every admin and placeholder avatar (the admin's initial, or a neutral users icon for placeholders) so a loading, slow, or broken image is never transparent. The whole greeting, including the "Hi" and "How can we help" headings, now waits for the avatars so the welcome animates in as one sequence. The 3s timeout safety net is kept, and a single broken image still unblocks the animation.

The redundant `Image()` preload, its `AbortController`, and `decode()` machinery are removed since the avatars' own load is now the source of truth.

Regression guard: not warranted, single-component greeting choreography where the gate measured a parallel preload instead of the rendered image; no other call site shares this pattern.

`bun scripts/static-checks.ts` on the changed component passes.

### DIFF
--- a/src/lib/components/customer-support/threads-overview.svelte
+++ b/src/lib/components/customer-support/threads-overview.svelte
@@ -108,8 +108,13 @@
 		}
 	});
 
-	// Track which specific image URLs have been preloaded (reactive Set)
-	let preloadedUrls = new SvelteSet<string>();
+	// Track which avatar image URLs have actually finished loading in the DOM.
+	// Populated by each <Avatar>'s onLoadingStatusChange below, so the gate
+	// reflects the *displayed* image. bits-ui runs its own internal load and
+	// keeps the <img> at display:none until it settles; gating on a separate
+	// preload let the stack animate in while the rendered image was still hidden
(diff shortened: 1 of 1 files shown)


stickerdaniel:
* **#618** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/618?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/618?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #619, opened 2026-06-30
fix(billing): return to app after closing billing portal

stickerdaniel:
The Stripe billing portal's back button sent users to useautumn.com instead of back into the app. `openBillingPortal` was called with no `returnUrl`, so Autumn used its org `success_url`, which defaults to `https://useautumn.com`.

Passing `returnUrl: window.location.href` at both manage-billing call sites (pricing cards, nav user menu) returns the user to the page they came from. Checkout already passes a `successUrl`, so only the portal was affected.

Regression guard: not warranted, two fixed call sites; reliably detecting a missing returnUrl needs type-aware lint, not a regex.

Spotted while QAing a live billing portal.

### DIFF
--- a/src/blocks/pricing/pricing-three.svelte
+++ b/src/blocks/pricing/pricing-three.svelte
@@ -87,7 +87,9 @@
 	}
 
 	async function handleManageBilling() {
-		await portalOperation.execute({});
+		// Return the user to the current page after the portal; without returnUrl
+		// Autumn falls back to its org success_url, which defaults to useautumn.com.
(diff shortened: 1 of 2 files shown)


stickerdaniel:
* **#619** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/619?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/619?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #627, opened 2026-07-03
fix(dev): Wipe local backend state before writing the auth secret

### LINKED ISSUE #625: RESET_LOCAL_BACKEND wipes the fresh better-auth secret, auth breaks on every following run
`RESET_LOCAL_BACKEND=true bun run dev` (or `test:e2e`) leaves the local backend in a state where every subsequent non-reset run fails all sign-ins.

Sequence:

1. `getPersistentBetterAuthSecret(cwd, reset=true, ...)` in vite.config.ts rolls a new secret and writes it to `.convex/<stateId>/better-auth-secret`.
2. convex-vite-plugin boots with `reset: resetLocalBackend` and runs `fs.rmSync(backendDir, { recursive: true, force: true })`, deleting the secret file that was just written.
3. The reset run itself works because the backend received the secret via env vars and keeps it in memory. Better Auth encrypts the new JWKS private key with this secret.
4. The next run finds no secret file, rolls a different secret, and mints tokens against the surviving JWKS rows. `/api/auth/convex/token` returns 500 with `BetterAuthError: Failed to decrypt private key. Make sure the secret currently in use is the same as the one used to encrypt the private key.` Every sign-in then redirects back to `/signin`.

Repro: `RESET_LOCAL_BACKEND=true bun run dev`, sign in (works), stop, `bun run dev`, sign in again (redirect loop, token endpoint 500s). Recovers only via another reset.

Fix that works for me: wipe the state dir in vite.config.ts right after reading `RESET_LOCAL_BACKEND`, before the secret file is written, and pass `reset: false` to convexLocal so the plugin never deletes the dir itself. The reset param of `getPersistentBetterAuthSecret` becomes redundant since the file is already gone after the wipe.

stickerdaniel:
Closes #625

RESET_LOCAL_BACKEND=true wrote the fresh better-auth secret into the state dir and then let convex-vite-plugin rm the whole dir at boot, deleting the file. The reset run kept working off the in-memory value, but every following run re-rolled the secret against the surviving JWKS rows, so /api/auth/convex/token returned 500 (BetterAuthError: Failed to decrypt private key) and every sign-in looped back to /signin until the next reset.

The config now wipes the state dir right after reading RESET_LOCAL_BACKEND, before the secret file is written, and passes reset: false to convexLocal so the plugin never deletes the dir itself. The reset parameter of getPersistentBetterAuthSecret is gone since the wipe already removed the file by the time it runs.

Regression guard: not warranted, the failure lives in the cross-process ordering between config evaluation and the plugin's boot-time rm, which unit tests can't observe; guarded manually by paired reset/plain boots.

bun scripts/static-checks.ts vite.config.ts passes. Verified with a reset boot followed by a plain boot: the secret file survives the reset run, the second run logs "Resuming backend from existing state" with the same secret, and the token endpoint keeps working.

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -39,20 +39,14 @@ function computeLocalConvexStateId(projectDir: string, suffix?: string): string 
 	return `${sanitizedBranch}${sanitizedSuffix}-${hash}`;
 }
 
-function getPersistentBetterAuthSecret(
-	projectDir: string,
-	reset: boolean,
-	suffix?: string
-): string {
+function getPersistentBetterAuthSecret(projectDir: string, suffix?: string): string {
 	const stateId = computeLocalConvexStateId(projectDir, suffix);
 	const stateDir = path.join(projectDir, '.convex', stateId);
 	const secretPath = path.join(stateDir, 'better-auth-secret');
 
-	if (!reset) {
-		const existing = fs.existsSync(secretPath) ? fs.readFileSync(secretPath, 'utf-8').trim() : '';
-		if (existing) {
-			return existing;
-		}
+	const existing = fs.existsSync(secretPath) ? fs.readFileSync(secretPath, 'utf-8').trim() : '';
+	if (existing) {
+		return existing;
 	}
 
 	const secret = crypto.randomBytes(32).toString('hex');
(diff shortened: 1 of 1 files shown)


stickerdaniel:
* **#627** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/627?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/627?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #628, opened 2026-07-03
fix(worktree): Reject extra args and warn on non-trunk base

stickerdaniel:
The worktree script read only the first positional, so an attempted positional base branch (bun run worktree fix/x main) was silently dropped and the new branch forked from whatever the invoking checkout happened to be on. Stacked on a feature branch, a later gt submit then submits that whole stack and puts unrelated WIP into the new PR. Mistyped flags were swallowed the same way because parsing ran non-strict.

Extra positionals now fail loudly with a pointer to --base, unknown flags are rejected via strict parseArgs, and when the implicit base (current branch) is not the trunk the script warns and suggests --base.

Regression guard: not warranted, the added arg guards are themselves the guard rails; no test harness exists for CLI arg parsing in scripts/.

bun scripts/static-checks.ts scripts/create-worktree.ts passes. Verified all three guards by invocation: extra positional and mistyped flag exit 1 before any git state is touched, and a run from a feature-branch checkout prints the trunk warning.

### DIFF
--- a/scripts/create-worktree.ts
+++ b/scripts/create-worktree.ts
@@ -21,22 +21,44 @@ const colors = {
 	red: '\x1b[31m'
 };
 
-// Parse command line arguments
-const { values, positionals } = parseArgs({
-	args: Bun.argv.slice(2),
-	options: {
-		'setup-only': { type: 'boolean', default: false },
-		'open-editor': { type: 'string' },
-		base: { type: 'string', short: 'b' },
-		help: { type: 'boolean', short: 'h', default: false }
-	},
-	strict: false,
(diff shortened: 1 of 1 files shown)


stickerdaniel:
* **#628** <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/628?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/stickerdaniel/saas-starter/628?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #698, opened 2026-07-12
docs(admin): make the admin doc a guide, not a plan

stickerdaniel:
`docs/admin-framework-plan.md` is 674 lines describing an admin framework that is already built: `src/lib/convex/admin/**` ships with tests, and the resource-driven panel it describes is what the template runs.

It is a good document wearing the wrong clothes. 605 of its lines are a genuine guide, with the architecture, the core concepts, the component and Convex layers, and resource registration with worked examples. That part earns its place and a fork author needs it to add a resource.

The last three sections do not. "What We Already Have vs. What to Build", "Migration Path" and "Implementation Priority" are plan, and a plan for something that shipped can only be wrong. They are wrong today, and every fork inherits them.

So: drop the three plan sections, keep the guide, and name the file what it is. The distinction is not cosmetic. A guide is a thing people maintain, because it describes what is. A plan is a thing people abandon at the merge, because it described what would be. Keeping a finished plan in the repo guarantees it rots, and the reader who trusts it wastes a day building something that exists.

Pairs with #697, which adds the rule this follows from: a plan is a PR body, not a repo file.

No inbound references to the old path anywhere in the repo, so nothing dangles.

### DIFF
--- a/docs/guides/admin-framework.md
+++ b/docs/guides/admin-framework.md
@@ -1,6 +1,8 @@
-# Admin Framework Plan
+# The admin framework
 
-Building a declarative, resource-driven admin panel inspired by Laravel Nova — implemented in Svelte 5 + TypeScript + Convex.
+A declarative, resource-driven admin panel inspired by Laravel Nova, implemented in Svelte 5 + TypeScript + Convex.
+
+This is a guide to the framework as it is built (`src/lib/convex/admin/**`, `src/lib/components/authenticated/`), not a plan for building it. Read it to add a resource, a field type, or a filter.
 
 ---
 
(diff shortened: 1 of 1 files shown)


stickerdaniel:
Closing: I was wrong. The framework is NOT built.

I inferred `src/lib/convex/admin/**` exists therefore the plan shipped. It does not follow. The plan's core API is a `defineResource` builder validating attribute keys against the Convex schema via `DocFields<TTable>`, and `defineResource` appears nowhere in `src/`. What exists under `admin/` is hand-written queries and mutations, not a declarative resource framework.

So the three sections I proposed deleting ("What We Already Have vs. What to Build", "Migration Path", "Implementation Priority") are the accurate part of the document, and this is a live plan for unbuilt work.

The reasoning that misled me is worth naming: I counted 62 guide-signals against 8 plan-signals and concluded it was a guide wearing a plan's clothes. But a detailed plan with code examples necessarily reads like a guide, because it describes how the thing should look. The signal count was a proxy. One grep for the plan's own core symbol would have settled it.

PR #784, opened 2026-08-07
fix(settings): Check the avatar bytes that actually upload

stickerdaniel:
Follow-up to #783, which moved the avatar type check to the wrong side of the transcoder.

The check runs on the picked file, but `downscaleImage` runs afterwards and re-encodes to WebP. Because it scales to 512px first, that re-encode is far smaller than a camera photo, so the conversion reliably wins the `blob.size < file.size` comparison. A HEIC picked in Safari therefore reached the server as a valid `image/webp` before #783 — and is now rejected up front. HEIC is the iOS camera default, so this is the normal case on iPhone.

The narrowing was still right in substance: `downscaleImage` is not total. It passes GIFs through, keeps the original when the re-encode is larger, and falls back to it on any decode failure, so an unsupported type can still arrive at the server. The mistake was checking the wrong bytes.

A profile now distinguishes what may be **stored** from what may be **picked**, which only differ where a transcoder sits in between. The avatar declares `image/*` as its input family, so the picker offers what the surface can genuinely take, and the type is verified again on the transcoder's output immediately before upload. Surfaces without a transcoder declare no input family and behave exactly as before — `acceptsSource` falls through to `acceptsMimeType` for them, which a test pins.

Regression guard: added `picking versus storing` in `src/lib/uploads/profiles.test.ts`

Verified with the unit suite and a project-wide type check. The new tests cover the distinction directly: HEIC and SVG are pickable but not storable, chat attachments answer identically to both questions, and the accept attribute offers the family rather than the stored extensions. One unrelated test (`prerender-sync`) timed out on a machine at load average 136 and passes on its own.

### DIFF
--- a/src/lib/uploads/profiles.test.ts
+++ b/src/lib/uploads/profiles.test.ts
@@ -2,16 +2,18 @@ import { describe, expect, it } from 'vitest';
 import {
 	acceptAttribute,
 	acceptsMimeType,
+	acceptsSource,
 	allowedMimeTypes,
 	UPLOAD_PROFILES,
+	type UploadProfile,
 	type UploadProfileName
 } from './profiles';
 import { canConsume, PROFILE_CONSUMERS, type Consumer } from './consumers';
 import { mediaCategoryOf, MODEL_INPUT_MODALITIES } from './model-capabilities';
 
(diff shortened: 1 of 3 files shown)


stickerdaniel:
Why not match picking surface with storing surface?

PR #837, opened 2026-08-24
fix(lint): Scan source for control bytes before checkers run

### LINKED ISSUE #832: Subprocess output can carry a control character from source into the terminal
`scripts/static-checks.ts` runs `prettier --check` before ESLint and prints child output verbatim. When a source file both contains a literal control character and fails to parse, Prettier's error includes a code frame with the original line, so the character reaches the terminal that prints the error. An ESC byte there is an ANSI escape, which the terminal executes.

Measured while reviewing #806: a valid comment containing U+001B produced no reproduction, while an invalid token containing U+001B produced a code frame carrying the character, and it survived stripping ANSI sequences from the output.

The new `local/no-literal-control-char` rule is safe by itself: it reports the codepoint and never quotes the source, and ESLint's stylish formatter does not print the offending line. Prettier runs first and never reaches that rule, so the pipeline still has the hole the rule was written to avoid.

Not fixed in #806 because the obvious fix is worse than the defect. Stripping ESC from child output would strip the colours ESLint and oxlint emit deliberately, and stripping only the dangerous subset means parsing another tool's escape grammar with a heuristic, which earns a false positive on every future round. It needs a decision about which streams go through a sanitiser and what CI output should look like afterwards.

stickerdaniel:
Closes #832
Regression guard: added `scripts/source-safety.test.ts`

Parsers can copy raw source into diagnostics before ESLint reports it safely. Reordering checks leaves other parsers and file types open, while output filtering cannot distinguish source-owned CSI from the identical bytes a formatter emits for colour.

The preflight runs before the static-check import graph, installation generators, formatting, and unit tests. It walks the working tree independently of Git's global ignores, validates every filename strictly, streams known text formats, fails closed on malformed UTF-8, and retains only the findings its bounded report can display. Generated email previews keep a narrow content exemption while their filenames remain covered. The ESLint rule, Svelte parser wrapper, and preflight share one character policy.

Verified with the static-check pipeline and complete unit suite. Twenty-seven cases cover parser leaks, malformed encoding, ignored files, filename structure, opaque assets, stream boundaries, report bounds, and collision-safe fixtures.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -39,7 +39,7 @@
 			"args": ["run", "static-checks:staged"],
 			"group": "test",
 			"problemMatcher": ["$eslint-compact", "$tsc"],
-			"detail": "Run pre-commit checks on staged files only",
+			"detail": "Scan source safety across the tree, then run staged pre-commit checks",
 			"presentation": {
 				"clear": true
 			}
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -57,7 +57,7 @@ The command fetches and branches from remote trunk by default. Worktrees live be
 
 Never use `EnterWorktree`. Prefix every later action with the absolute worktree path because command working directories do not persist. Before committing, run `git branch` and confirm the worktree branch. After a merge, run `bun run worktree:prune`; it safely removes confirmed-merged worktrees and branches and fast-forwards a clean local trunk.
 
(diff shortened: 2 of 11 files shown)


stickerdaniel:
Superseded by #847. Review showed that an in-repository source preflight cannot protect the package manifest or its own bootstrap module before a parser reads them. The replacement puts the boundary at rendered checker output instead.

PR #877, opened 2026-08-31
fix(chat): Stream clause deltas instead of whole lines

stickerdaniel:
A streamed reply stalls before it moves. Line chunking buffers text until a newline arrives, so a paragraph of prose produces no delta at all while the model is generating it, and the reader watches a still screen until the paragraph lands whole. Measured on a three-paragraph answer at 45 tokens/s, 1244ms passed with no update; on a single paragraph containing no newline, nothing is written until the reply ends. The presentation scheduler on the base branch paces the words inside a burst, but it cannot fill the silence before one.

Both call sites now share one constant that names the agent's own default. Omitting the option would not reach it, because the agent forwards `undefined` to the AI SDK's `smoothStream`, whose default is word chunking instead. The cost is more writes, and it is bounded by the throttle rather than by the chunk size: `DeltaStreamer.addParts` coalesces everything arriving inside one window, so the same answer costs 30 writes rather than 2 and can never exceed ten per second of generation. The guard measures the stall on newline-free prose instead of asserting the value, since the value alone says nothing about which setting stalls.

Two suites had replaced all of `@convex-dev/agent` with the calls they assert on, which failed them at load as soon as anything new was imported from it; they now use the `importOriginal` shape a third suite already used. Verified with the full unit suite, the Convex type check, and the compat check.


stickerdaniel:
Blocked on a defect this change would expose, found in cross-model review and confirmed against the sources.

The consumer requests deltas from cursor zero on every subscription update (`src/lib/chat/ui/ChatRoot.svelte:148`), and the agent component returns at most `MAX_DELTAS_PER_STREAM = 100` documents per stream (`@convex-dev/agent/src/component/streams.ts:31,67-74`). Nothing merges delta documents: `compressUIMessageChunks` only concatenates parts inside one delta, so document count equals `addDelta` count. During a stream the rendered text comes solely from those deltas (`deriveUIMessagesFromDeltas` via `src/lib/chat/ui/streaming-display.ts:54`), with no second path supplying it until the final message is saved.

At one delta per 100ms throttle window, a reply that generates for more than ten seconds exceeds 100 deltas; from there the query keeps returning the same first 100, the visible reply freezes, and the text appears in full only when generation ends. Line chunking rarely reached that count, so the limit is pre-existing but this change makes it the ordinary case for long replies.

Holding this until the cursor is advanced and pages accumulated, which is its own change and needs its own review.

stickerdaniel:
Measured, not only read. Wrote 150 deltas straight through the component into one stream, then read them back on the path `ChatRoot` uses (`listDeltas` with `cursor: 0`):

```
written:  150
returned: 100
lastEnd:  796   # delta 99 of 150
```

The query returns exactly 100 regardless of how many exist, so the rendered text stops at the 100th delta and only completes when the final message is saved. Confirms the finding above.

